PR #6, opened 2019-06-09
Rework autotools build system

rw-anderson:
This branch reworks the automake system for librom.  It attempts to simplify, clean up, standardize, and add a few new features, like building a shared library in addition to the static library.

This was motivated by trying to build librom from spack, which ran into a bunch of problems with the automake system (automake version mismatches, compiler flags that shouldn't be there, broken make dist target, etc.)  Spelunking into the casc m4 macros showed special cases for systems which haven't been around for over a decade, so it looked like a refresh was in order.  I tried to use autoconf library standard macros where possible.  This may make things a little less automatic for LLNL users, but should be more general, to some approximation.  The shared library feature was motivated by playing around with the possibility of creating jupyter notebook(s) to demonstrate usage of librom. jupyter notebooks using cling can only interact with shared libraries.

**This will need testing on multiple platforms before merging.**

Note that for reviewing these changes, most of the changes are not particularly relevant as they are autogenerated files, or the removal or addition of library supplied macros.  The files that comprise the substance of the changes are:

Makefile.am
configure.ac
README
autogen.sh

autogen.sh is for re-running autotools.  This can be run on a fresh checkout.  I think ultimately all of the autogenerated files should be removed from the git repo, and we can use make dist to supply tarballs that don't depend on autotools, but contain generated configure and Makefile.in.

There is an associated spack branch that will need a pull request once this is merged.
https://github.com/rw-anderson/spack/tree/features/rwa/librom-update


.gitignore
```diff
@@ -1,10 +1,30 @@
-# Do not track objects or archives
```
goxberry: This line shouldn't need to be changed. I can add the commit reverting this change, if that's okay.

README.md
```diff
@@ -59,6 +59,19 @@ See the INSTALL file for details; in brief, it uses the `configure`,
...
+An example configuration line that picks up packages installed at
+various paths might look like the following.  Note that depending on
+the package, the --with-<pkg>= directive expects slightly different
+things.  hdf5 wants a compiler wrapper; blas and lapack want lists of
+linker flags; elemental and gtest want an installation prefix.
+
+./configure --with-CXX=<mpipath>/bin/mpic++ \
+            --with-hdf5=<hdf5path>/bin/h5cc \
+	    --with-blas='-L<blaspath> -lopenblas' \
+	    --with-lapack='-L<blaspatch> -lopenblas' \
```
goxberry: "blaspatch" -> "lapackpath"

LAPACK doesn't necessarily have to share the BLAS path; Netlib's BLAS could be installed in a separate path from Netlib's LAPACK, for instance, or [BLIS](https://github.com/flame/blis) could be used with Netlib's LAPACK or the LAPACK provided by [libflame](https://github.com/flame/libflame). But some common LAPACK implementations also distribute a BLAS implementation (Netlib's LAPACK, OpenBLAS, MKL).

README.md
```diff
@@ -59,6 +59,19 @@ See the INSTALL file for details; in brief, it uses the `configure`,
...
+	    --with-elemental=<elpath> \
```
goxberry: Given PR #12, it's probably time to strip out the [Elemental](https://github.com/elemental/Elemental) support -- Elemental is deprecated and the LLNL fork strips out all functionality except for BLAS Level 1 subroutines. We don't need to make this change in this PR -- I'll add a separate issue.

README.md
```diff
@@ -59,6 +59,19 @@ See the INSTALL file for details; in brief, it uses the `configure`,
...
+	    --with-gtest=<gtestpath>
+
 # For developers who need to hack the build system
 
 If you want to add new source files and executables, the best way to
```
goxberry: The line containing `autoreconf -I config --install` probably needs to be changed to `autoreconf --install` (line 77 of the version currently in `master`, line 90 of the version in this PR).

m4/README
```diff
@@ -0,0 +1,4 @@
+m4 macros needed by automake system.
+
+ax_require_defined.m4: needed by ax_cxx_compile_stdcxx.m4
+
```
goxberry: Should there be more files from the [Autoconf Archive](https://www.gnu.org/software/autoconf-archive/) listed here?

travis_ci/README
```diff
@@ -0,0 +1,3 @@
+
+These are helper scripts potentially useful in the travis
```
goxberry: "These are helper scripts" -> "This directory contains helper scripts"

goxberry:
So I've run into two issues on my MacBook Pro:

1) I still run into version issues with the GNU Autotools toolchain. If I have autoconf 1.16.1, the build system throws an error that it can't locate `aclocal` from autoconf 1.13. A similar issue arises with libtool.

2) When I use shared libraries, I'm encountering a dynamic linking error that could be fixed if every flag of the form `-L${SOME_PATH}` were accompanied by a flag of the form `-Wl,-rpath,${SOME_PATH}` (note that the equivalent `-Wl,-rpath=${SOME_PATH}` will work on Linux, but not macOS). I'm going to see if this issue is fixable another way.

.gitignore
```diff
@@ -1,10 +1,30 @@
...
+m4/lt~obsolete.m4
+compile
+libtool
+ltmain.sh
+ar-lib
+.deps
+.libs
+.dirstamp
+
+# Do not track emacs backup files
```
goxberry: These editor-specific settings should be in a user's "global" `.gitignore` file, specified by the `core.excludesFile` variable in `$HOME/.gitconfig` (if you want these patterns to apply to all of the git repos you work on), or in `$GIT_DIR/info/exclude` (if you these patterns to apply only to this git repo) -- neither file should be tracked in the repo. (See the second bulleted list under "Description" in https://git-scm.com/docs/gitignore.) Delegating editor-specific `.gitignore` settings to these files eliminates the possibility of accumulating pattern cruft from multiple editors or IDEs (e.g., Sublime Text, Atom, TextMate, vim, nano, CLion, Eclipse, NetBeans, Visual Studio Code).

goxberry:
A decent chunk of this PR could be replaced by #20, if there is interest. Depending on how atomic the PR commits are, it might be possible to cherry-pick some commits over. Otherwise, some repo surgery (`git-rebase -i`) or hackery (liberal use of `git-show` with stdout redirected to files) will be needed to migrate the Travis CI configuration over. In either case, I'd like to preserve @rw-anderson's authorship of those commits, which could be done with `git-commit --author=${AUTHOR_INFO} --amend`, if neither the `git-rebase` nor the `git-show` approaches yield the desired authorship for the commits in question.

travis_ci/configure-for-travis-vm
```diff
@@ -0,0 +1,22 @@
...
+#
+# and also that googletest has been installed via spack.
+
+gtestdir=$( spack find -p googletest | tail -1 | awk '{print $2}' )
+
+./configure \
+	    --with-CXX=/usr/bin/mpic++ \
+	    --with-hdf5=/usr/bin/h5cc \
+	    --with-lapack='-lopenblas' \
+	    --with-gtest=${gtestdir}
```
goxberry: In light of some of the memory bugs I'm seeing, it may be wise to also add `--enable-opt=yes \ CXXFLAGS="-fsanitize=address"` (or similar). [Address Sanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer)(ASan) can be used to detect memory leaks, even on optimized code. I wouldn't recommend running ASan on code with compiler optimization flags in a debugger when trying to find and fix a memory leak (unless, of course, the bug does not occur when running with `-O0`), but in this case, we don't care about stepping through the code -- we only care whether or not memory leaks are present when running the tests on a CI server. ASan is available in GCC as of version 4.8 and in LLVM-based compilers as of LLVM version 3.1, so it should be possible to run our tests with ASan on Travis CI. ASan should slow execution by ~2x and can be run on optimized code, as opposed to Valgrind/memcheck, which slows execution by ~20x and its authors recommend not using it on optimized code. Hopefully, this performance penalty doesn't slow down execution to the point where testing exceeds Travis CI limits; otherwise, we'll have to figure out workarounds.

m4/ax_lapack.m4
```diff
@@ -0,0 +1,131 @@
...
+AC_DEFUN([AX_LAPACK], [
+AC_REQUIRE([AX_BLAS])
+ax_lapack_ok=no
+
+AC_ARG_WITH(lapack,
+        [AS_HELP_STRING([--with-lapack=<lib>], [use LAPACK library <lib>])])
+case $with_lapack in
+        yes | "") ;;
+        no) ax_lapack_ok=disable ;;
+        -* | */* | *.a | *.so | *.so.* | *.o) LAPACK_LIBS="$with_lapack" ;;
```
goxberry: There's a bug for macOS in this line -- macOS dynamic libraries take the form `*.dylib` and `*.dylib.*`, not `*.so` and `*.so.*` as in Linux/Solaris/*BSD. This issue may explain why dynamic library linking wasn't working on my Mac. I'll be submitting a PR with a fix to https://github.com/autoconf-archive/autoconf-archive shortly.

goxberry: I fixed a similar issue in hypre a few months ago, which is the only reason I even thought to look for this sort of issue.

m4/ax_blas.m4
```diff
@@ -0,0 +1,238 @@
...
+AC_REQUIRE([AC_F77_LIBRARY_LDFLAGS])
+AC_REQUIRE([AC_CANONICAL_HOST])
+ax_blas_ok=no
+
+AC_ARG_WITH(blas,
+	[AS_HELP_STRING([--with-blas=<lib>], [use BLAS library <lib>])])
+case $with_blas in
+	yes | "") ;;
+	no) ax_blas_ok=disable ;;
+	-* | */* | *.a | *.so | *.so.* | *.o) BLAS_LIBS="$with_blas" ;;
```
goxberry: As with the LAPACK macro, this collection of globs again omits the macOS-specific dynamic library globs `*.dylib` and `*.dylib.*`. Looks like I'll have to file another PR with a fix to https://github.com/autoconf-archive/autoconf-archive.

goxberry: Submitted https://github.com/autoconf-archive/autoconf-archive/pull/194.

m4/ax_lapack.m4
goxberry: Submitted https://github.com/autoconf-archive/autoconf-archive/pull/194.

goxberry:
Here's one of the errors I see -- something about a `libtool` version mismatch:

```console
> # Run configure, then...
> make
/Applications/Xcode.app/Contents/Developer/usr/bin/make  all-am
depbase=`echo Matrix.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
	/bin/sh ./libtool  --tag=CXX   --mode=compile /usr/local/opt/open-mpi/bin/mpicxx -std=c++11 -DHAVE_CONFIG_H -I.  -I. -I/Users/oxberry1/repos/spack/opt/spack/darwin-mojave-x86_64/clang-10.0.1-apple/googletest-1.8.1-fntx3sosfxgmlkusxxrgtptdc2cckhdv/include  -I/usr/local/Cellar/hdf5/1.10.5_1/include -I/usr/local/opt/szip/include  -g -g -O2 -MT Matrix.lo -MD -MP -MF $depbase.Tpo -c -o Matrix.lo Matrix.C &&\
	mv -f $depbase.Tpo $depbase.Plo
libtool: Version mismatch error.  This is libtool 2.4.6, but the
libtool: definition of this LT_INIT comes from libtool 2.4.2.
libtool: You should recreate aclocal.m4 with macros from libtool 2.4.6
libtool: and run autoconf again.
make[1]: *** [Matrix.lo] Error 63
make: *** [all] Error 2
```

I can fix this issue by regenerating the configure script, but my impression was that we wanted to avoid doing so.

goxberry:
I've narrowed the problem down to libtool. Whatever libtool is doing, even though I've added `-Wl,-rpath,${GTEST_PREFIX}/lib` to the compile/link flags, in addition to all of the other flags, whenever I run one of the test executables on my LLNL MacBook Pro, I get

```console
> ./tests/test_Vector
dyld: Library not loaded: libgtest.dylib
  Referenced from: /Users/oxberry1/repos/libROM/tests/.libs/test_Vector
  Reason: image not found
  Abort trap: 6
```

goxberry:
After careful consideration, I've decided to close this pull request in favor of #20. The main factors driving this decision have been:

- as implemented, rebuilding the GNU Autotools build system returns warnings or errors if the local versions of, e.g., autoconf and libtool differ from the versions used to generate the configure script
- these issues haven't been fixed over the life of the project, specifically over the last 6 to 8 months
- kludging around these issues has led to a lot of noise in git commits
- internally at LLNL due to [BLT](https://github.com/LLNL/blt) and externally in the DOE software ecosystem (e.g., the [xSDK](https://xsdk.info/)), CMake has gained greater traction
- rebuilding the CMake build system is not nearly as fragile if the local CMake version differs from the CMake version used in developing the CMake script; deprecation policies seem more graceful
- recently, multiple LLNL collaborators have mentioned issues in compiling and installing libROM with the current build system, suggesting that an urgent change is needed

If there is some way to rescue this PR at a future date, I think it can be reopened or salvaged.

PR #8, opened 2019-06-19
Matrix bugfix: include header with FC_FUNC macros

goxberry:
As of this PR, `master:Matrix.C` uses the `FC_FUNC` macros without including `CAROM_config.h`, so these macros are sometimes undefined (depending on order of execution of commands). This commit includes this header in `Matrix.C` so that the `FC_FUNC` macros used for Fortran name-mangling are defined, fixing compile-time errors in some cases.

### DIFF
--- a/Matrix.C
+++ b/Matrix.C
@@ -23,6 +23,7 @@
 #endif
 
 /* Use Autotools-detected Fortran name-mangling scheme */
+#include "CAROM_config.h"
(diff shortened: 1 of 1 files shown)


goxberry:
Rebased on `master` after merging #12.

goxberry:
Merging, because it's a one line change already discussed at our progress meeting on 2019-07-10, and this change does not conflict with #6.

PR #9, opened 2019-06-19
[feature/matrix-add-vector-outer-product] Add a vector-vector outer product returning a Matrix

goxberry:
This pull request adds a function to compute a `CAROM::Matrix` from the outer product of two `CAROM::Vector` objects. This function does not require access to any private data members of `CAROM::Vector` objects, so it is a free function -- not a member function. I would like to keep the interfaces of our classes as small as possible.

### DIFF
--- a/Matrix.C
+++ b/Matrix.C
@@ -17,6 +17,7 @@
 
 #include "mpi.h"
 #include <string.h>
+#include <vector>
 
 #ifdef CAROM_HAS_ELEMENTAL
 #include <El.hpp>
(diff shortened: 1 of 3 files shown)


goxberry:
The general plan for setting up the testing infrastructure for the incremental SVD is as follows:

[ ] write a vector-vector outer product (this PR)

[ ] write a routine that computes Householder reflectors (another PR)

[ ] write a routine that makes a diagonal matrix from a `Vector`

[ ] write a routine that generates a random orthogonal matrix (with uniform probability, where the probability measure is the Haar measure over the orthogonal group)

After these tasks are done, we have enough infrastructure to test the SVD and QR decomposition algorithms in serial or on small numbers of parallel processes, so I can then write tests for the incremental SVD and add the singular value tests from the recent Fareed-Singler preprint.

goxberry:
More tasks:

- [  ] Add a serial incremental SVD test.

- [  ] Modify the incremental SVD to use the Fareed-Singler algorithm.

- [  ] Add infrastructure for parallel unit testing.

- [  ] Add a parallel-parallel matrix (distributed-distributed?) product.

- [  ] Add a parallel incremental SVD test.

goxberry:
Rebased on `master` after merging #12, then #8.

goxberry:
cc: @slmcbane

Matrix.C
```diff
@@ -1262,5 +1263,104 @@ const
...
+        for (int i = 0; i < num_cols_on_proc.size(); i++)
+        {
+            gathered_w_displacements.at(i) = result_num_cols;
+            result_num_cols += num_cols_on_proc.at(i);
+        }
+
+        // Gather the data from each process onto each process -- this
+        // step is an Allgatherv (because difference processes may
+        // have different numbers of entries.
+        std::vector<double> w_data(process_local_num_cols);
```
slmcbane: It seems like it would be nice to have a public `const double* Vector::data()` function to eliminate some copies like this. As a bonus, since LAPACK and BLAS are required anyway, we could then use them to do operations like an outer product faster than the for-loop implementation.

Issue #18, opened 2019-07-05
Remove Perl dependency.


goxberry:
If we're using GNU Autotools or CMake, we shouldn't need Perl anymore. (I believe Doxygen requires Perl, but Perl is no longer a direct dependency of this project.) The Perl dependency was introduced by the `depend.pl` script, but this script is no longer needed.

PR #26, opened 2019-07-17
[feature/remove-perl-dependency] Remove all references to Perl from the build system and project

goxberry:
Our build system currently requires a path to a Perl binary, but our project does not need Perl:

- Doxygen no longer requires Perl as of version 1.8.0, released on 2012-02-25. Most Linux clusters at LLNL have an operating system no more than 5 or 6 years old, and thus should have Doxygen 1.8.0 or greater installed (if installed at all); Quartz is one example.
- The Perl scripts in `scripts` are unused -- they generate dependency information (`depend` and `depend.pl`) or compare revision hashes in files (`cmp`). In the case of the former, we have two build systems that figure out dependency information for us -- the current GNU Autotools build system (plus changes in #6), and the proposed new CMake build system (see #15) -- and even if we didn't, that information could be generated automatically in a GNU Makefile using compiler invocations with appropriate flags (see http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/ for a good tutorial on how to do so). There's no need to involve Perl for dependency detection. As for revision hashes, we don't store them in files.

Given that our project does not need Perl, this PR sets about removing all traces of Perl from the project. This cleanup should also help with making the GNU Autotools and CMake build systems consistent (fewer macros to worry about).


goxberry:
Fixes #18.

PR #29, opened 2019-09-09
WIP: Add parametric ROM capability: save snapshots, load snapshots, and load basis vectors

jtlau:
This PR adds basic capabilities to support parametric ROM building. 

1. Save snapshots using the function **writeSnapshots()**
2. Load either basis vectors or snapshots using **loadSamples()**

There are two examples included:

1. **smoke_static.C** 
- Basic example showing use of functions 1 and 2 above.

2. **load_samples.C** 
- Example assuming snapshots or basis vectors are already saved from a full order model. This is the code needed in order to merge those snapshots/basis vectors and compute the SVD over them.

Future work:
- Code is written and organized in a way where this can be extended to work with the Incremental SVD algorithm by completing the **getSnapshotMatrix()** function (currently exists as a placeholder in IncrementalSVD.C). 
- The time is defaulted to 0.0 in the loadSamples() function since it isn't used for static SVD. This will also need to be changed (read from file) if extending to work with incremental SVD.


goxberry:
Is there some way we could simplify this PR? It looks like the whole ScalaWRAP project has been included, along with a large number of Doxygen-generated files. If we could get rid of those files -- generated files usually aren't tracked in a Git repo -- I think the number of files changed would decrease to a more manageable number.

Even better would be to figure out a way to avoid committing ScalaWRAP to this repo, if there's time for that.

jtlau:
Hi @goxberry - I think removing the Oxygen-Generated files and Scalawrap from this PR is doable. I will start doing so.

PR #31, opened 2019-11-26
Add parametric ROM capability 

jtlau:
This PR adds basic capabilities to support parametric ROM building. This is a cleaned-up version of my previous PR for the same capability. It was easier to create a new branch in my fork and then create a new PR. 

1. Save snapshots using the function **writeSnapshots()**
2. Load either basis vectors or snapshots using **loadSamples()**

There are two examples included:

1. **smoke_static.C**

- Basic example showing use of functions 1 and 2 above.

2. **load_samples.C**

- Example assuming snapshots or basis vectors are already saved from a full order model. This is the code needed in order to merge those snapshots/basis vectors and compute the SVD over them.

Future work:

- Code is written and organized in a way where this can be extended to work with the Incremental SVD algorithm by completing the **getSnapshotMatrix()** function (currently exists as a placeholder in IncrementalSVD.C).

- The time is defaulted to 0.0 in the loadSamples() function since it isn't used for static SVD. This will also need to be changed (read from file) if extending to work with incremental SVD.


goxberry:
@jtlau Is it all right if I make changes to this branch to bring it up to date with changes to `master`? I probably won't get to making changes until at least this coming weekend, but I'd like to get this functionality into libROM before the `master` branch diverges too much from your PR.

jtlau:
@goxberry yes, please make any changes you need. Thanks!

PR #33, opened 2020-06-22
MACchanges

kevinhkhuynh:
Hi, Youngsoo. These changes allow the library to work easily with MAC.

It will require more changes in user2.mk in Laghos though, but I think it simplifies user2.mk in Laghos.

This is my user2.mk now:

laghos: $(OBJECT_FILES) $(CONFIG_MK) $(MFEM_LIB_FILE) $(CCC) -o laghos $(OBJECT_FILES) $(LIBS) -Wl,-rpath,$(CURDIR)/../libROM/build -L$(CURDIR)/../libROM/build -lROM $(SCALAPACK_FLAGS)

I have added -rpath for libROM and removed hdf5 as we will use the hdf5 version libROM points to.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,7 @@ list(APPEND source_files
   scalapack_c_wrapper.c
   scalapack_f_wrapper.f90)
 
-add_library(ROM ${source_files})
+add_library(ROM SHARED ${source_files})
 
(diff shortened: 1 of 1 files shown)


kevinhkhuynh:
The Laghos regression test script also will not run correctly without a small change in that file as well.

PR #36, opened 2020-07-09
add fixes to compile with ardra

kevinhkhuynh:
toolchain file to compile with ardra

### DIFF
--- a/StaticSVDBasisGenerator.C
+++ b/StaticSVDBasisGenerator.C
@@ -20,6 +20,7 @@ StaticSVDBasisGenerator::StaticSVDBasisGenerator(
    int dim,
    int samples_per_time_interval,
    const std::string& basis_file_name,
+   bool output_rightSV,
    int max_basis_dimension,
    double sigma_tolerance,
    bool debug_algorithm,
@@ -30,7 +31,8 @@ StaticSVDBasisGenerator::StaticSVDBasisGenerator(
                                            samples_per_time_interval,
                                            max_basis_dimension,
                                            sigma_tolerance,
-                                           debug_algorithm));
+                                           debug_algorithm,
+                                           output_rightSV));
 }
 
 StaticSVDBasisGenerator::~StaticSVDBasisGenerator()
--- a/INSTALL
+++ b/INSTALL
@@ -1,117 +1,15 @@
 # Installing libROM
 
-1) Configure libROM by running:
+For LLNL LC Machines:
 
-```console
-./configure --with-CXX=$PATH_TO_MPICXX --with-lapack=$LAPACK_INSTALL_PREFIX \
---with-lapack-libs=$LAPACK_LINK_FLAGS_WITH_LITTLE_L_ONLY \
---with-hdf5=$HDF5_INSTALL_PREFIX \
---with-perl=$PERL_INSTALL_PREFIX \
---with-elemental=$ELEMENTAL_INSTALL_PREFIX \
---with-gtest=$GOOGLETEST_INSTALL_PREFIX \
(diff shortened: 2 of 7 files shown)


StaticSVDBasisGenerator.C
```diff
@@ -30,7 +31,8 @@ StaticSVDBasisGenerator::StaticSVDBasisGenerator(
                                            samples_per_time_interval,
                                            max_basis_dimension,
                                            sigma_tolerance,
-                                           debug_algorithm));
+                                           debug_algorithm,
+                                           output_rightSV));
```
chldkdtn, line 35: @kevinhkhuynh, Introducing output_rightSV is good, but it does not complete the task of enabling temporal basis write for "static" SVD, yet. Is it still in progress?

kevinhkhuynh: Hi Youngsoo. If the user sets output_rightSV to true, temporal basis is outputted. Otherwise, output_rightSV defaults to false and nothing is outputted. Temporal basis was already there, but wasn't being outputted before because in StaticSVDSampler.h, updateRightSV defaults to false, and the user never had a way to set it to true, but now they can with output_rightSV.

kevinhkhuynh:
> Is there a way of not having two separate versions, i.e., one for Ardra and another for Laghos? It would be nice we can use only one configure that works for both Ardra and Laghos so that one build of libROM works for both Ardra and Laghos.

As of now, I don't know if it would be possible without changing how we compile Laghos and, by extension, MFEM and the other dependencies.

kevinhkhuynh:
> > Is there a way of not having two separate versions, i.e., one for Ardra and another for Laghos? It would be nice we can use only one configure that works for both Ardra and Laghos so that one build of libROM works for both Ardra and Laghos.
> 
> As of now, I don't know if it would be possible without changing how we compile Laghos and, by extension, MFEM and the other dependencies.

If we want something slightly cleaner, we can have a single libROM directory with two build directories inside, such as build and buildArdra. So these two builds for Ardra and Laghos would use the same cpp/header include files.

kevinhkhuynh:
> > 
> 
> If we want something slightly cleaner, we can have a single libROM directory with two build directories inside, such as build and buildArdra. So these two builds for Ardra and Laghos would use the same cpp/header include files.

If we do it this way, it would require one small path change in space_time_rom_compile-fix

chldkdtn:
> > > 
> > 
> > 
> > If we want something slightly cleaner, we can have a single libROM directory with two build directories inside, such as build and buildArdra. So these two builds for Ardra and Laghos would use the same cpp/header include files.
> 
> If we do it this way, it would require one small path change in space_time_rom_compile-fix

Let's do this.

kevinhkhuynh:
I think it is better if the user creates buildArdra then leaving it in the repo.

chldkdtn:
> I think it is better if the user creates buildArdra then leaving it in the repo.

Thank you for doing this, Kevin. 
Can you also check "INSTALL" file in libROM home directory and check what is working and what is not? I think it is somewhat outdated.

Also, Travis CI for libROM is set at 

      https://travis-ci.com/LLNL/libROM

Can you check if that is running without any error?

kevinhkhuynh:
> > I think it is better if the user creates buildArdra then leaving it in the repo.
> 
> Thank you for doing this, Kevin.
> Can you also check "INSTALL" file in libROM home directory and check what is working and what is not? I think it is somewhat outdated.
> 
> Also, Travis CI for libROM is set at
> 
> ```
>   https://travis-ci.com/LLNL/libROM
> ```
> 
> Can you check if that is running without any error?

I have fixed it and put it on a new pull request.

PR #37, opened 2020-07-21
travis integration with cmake

kevinhkhuynh:
Travis Integration with CMake. CI testing runs on Mac and Linux. The passing results of the first build are shown in this URL:

https://travis-ci.org/github/LLNL/libROM/branches

They can also be seen down below near the merge button.


kevinhkhuynh:
Any currently checked out branches won't have TravisCI working, but any new branches that are checked out from master from now on will. This is because the old branches don't have .travis.yml in the repo and so TravisCI doesn't know to build on those branches.

PR #38, opened 2020-07-24
Filter small sv

chldkdtn:
This PR introduces Lines 17 to 19 in Algorithm 1 of the following paper:

H. Fareed and J. Singler, "Error analysis of an incremental proper orthorgonal decomposition algorithm for PDE simulation data," Journal of Computational and Applied Mathematics, 368 (2020),

which is the routine of filtering the small singular value.


kevinhkhuynh:
I think this merge might break Laghos, if we look at line 62 and 72 in the following file, the parameter for singular_value_tol isn't added in yet. So, Laghos and any other files that instantiate a IncrementalSVDBasisGenerator will also need to be updated.

https://github.com/CEED/Laghos/blob/rom-dev/laghos_rom.hpp

kevinhkhuynh:
I think this is the best we can do since deleting columns is not possible in C++ since the arrays are stored in row-order.

IncrementalSVDFastUpdate.C
```diff
@@ -171,6 +174,44 @@ IncrementalSVDFastUpdate::computeBasis()
...
    }
+
+  if(d_rank == 0) {
+       std::cout << "d_num_samples = " << d_num_samples << "\n";
+       std::cout << "d_num_rows_of_W = " << d_num_rows_of_W << "\n";
+       std::cout << "d_singular_value_tol = " << d_singular_value_tol << "\n";
+       std::cout << "smallest SV = " << d_S->item(d_num_samples-1,d_num_samples-1) << "\n";
+       std::cout << "next smallest SV = " << d_S->item(d_num_samples-2,d_num_samples-2) << "\n";
+   }
+   // remove the smallest singular value if it is smaller than d_singular_value_tol
```
kevinhkhuynh, line 185: I have a question. Is computeBasis called multiple times, at every iteration of the IncrementalSVD? Or is it only called once by buildInitialSVD?

chldkdtn: computeBasis is also called in "buildIncrementalSVD" in IncrementalSVD.C. It is supposed to be called whenever the incremental basis is updated.

chldkdtn:
> I think this is the best we can do since deleting columns is not possible in C++ since the arrays are stored in row-order.

which is somewhat unfortunate. :(

chldkdtn:
> I think this merge might break Laghos, if we look at line 62 and 72 in the following file, the parameter for singular_value_tol isn't added in yet. So, Laghos and any other files that instantiate a IncrementalSVDBasisGenerator will also need to be updated.
> 
> https://github.com/CEED/Laghos/blob/rom-dev/laghos_rom.hpp

It totally breaks Laghos ROM. Right now, it seems that it breaks ardra STROM simulations also. Once the ardra strom issue is resolved, we will need to update Laghos ROM.

kevinhkhuynh:
updated with master and ready for review.

chldkdtn:
> updated with master and ready for review.

@kevinhkhuynh You can merge this branch to master when it is appropriate.

PR #39, opened 2020-08-07
Maxtimeinterval

kevinhkhuynh:
1. Added option to add a maximum time interval. Defaults to -1 (unlimited time intervals). If the number is surpassed (more samples are taken than is allowed), libROM throws an error.
2. Added tests for this.
3. Modified travis to run the tests, before it just built them, my mistake.
4. Fixed Cmake bug where Debug mode didn't turn on assertions.
5. Verified that Laghos regression tests still pass and Dylan's failed example actually fails.


kevinhkhuynh:
need to fix travis yml

PR #40, opened 2020-08-07
Add max time intervals

kevinhkhuynh:
1. Added option to add a maximum time interval. Defaults to -1 (unlimited time intervals). If the number is surpassed (more samples are taken than is allowed), libROM throws an error.
2. Modified travis to run the tests, before it just built the test executables, my mistake.
3. Fixed Cmake bug where Debug mode didn't turn on assertions.
4. Verified that Laghos regression tests still pass and Dylan's failed example actually fails.


kevinhkhuynh:
So if we build libROM with gtest and run ./test_Matrix, it seems MatrixSerialTest.Test_qrcp_pivots_transpose passes sometimes, but other times it doesn't, almost randomly, like some type of error due to parallelization. I didn't modify this file so it seems this error has already been here.

Matrix.C
```diff
@@ -842,7 +842,7 @@ Matrix::qrcp_pivots_transpose_serial(int* row_pivot,
   double* work = new double[lwork];
   double* tau  = new double[std::min(num_rows_of_transpose,
 				     num_cols_of_transpose)];
-  int* pivot = new int[num_cols_of_transpose];
+  int* pivot = new int[num_cols_of_transpose] ();
```
kevinhkhuynh, line 845: So before, this was randomly initialized. Depending on its initialization, dgeqp3 was returning a different answer. I initialized it to all zeroes and the test passes. I'm not sure if this is correct though.

http://www.netlib.org/lapack/explore-html/dd/d9a/group__double_g_ecomputational_ga1b0500f49e03d2771b797c6e88adabbb.html

SVDBasisGenerator.h
```diff
@@ -107,13 +107,9 @@ class SVDBasisGenerator
          if (getNumBasisTimeIntervals() > 0 &&
              d_svdsampler->isNewTimeInterval()) {
             d_svdsampler->resetDt(dt);
-            // YC: commenting this out unless someone think this is necessary
-            //     we call writeBasis in "endSamples()" function below.
```
kevinhkhuynh, line 111: It seems that the functionality was here, we just needed to uncomment this.

kevinhkhuynh: I believe that BasisWriter.C puts everything into one hdf5 file, but puts each interval in its own dataset, like the following: sprintf(tmp, "spatial_basis_%06d", d_num_intervals_written); where the variable represents the current time interval.

kevinhkhuynh: I looked into basisWriter->writeBasis. It calls "getSpatialBasis" within, which runs the computeSVD. So we would only need to take samples, with this automatically writing out the basis for every interval, except the last, which "endSamples" will cover since it also calls writeBasis.

SVDBasisGenerator.h
```diff
@@ -108,7 +108,8 @@ class SVDBasisGenerator
              d_svdsampler->isNewTimeInterval()) {
             d_svdsampler->resetDt(dt);
             if (d_basis_writer) {
-                d_basis_writer->writeBasis();
+                d_basis_writer->writeBasis("basis");
+                writeSnapshot();
```
kevinhkhuynh, line 112: I've added this line to write the snapshot matrix as well whenever a time interval ends. It doesn't do the last one though, so the user will still have to call writeSnapshot for the last one, similar to calling endSamples. Do we want this?

IncrementalSVD.h
```diff
@@ -40,6 +40,7 @@ namespace CAROM {
        *                                    samples.
        * @param[in] samples_per_time_interval The number of samples to be
        *                                      collected for each time interval.
+       * @param[in] max_time_intervals The maximium number of time intervals.
```
dylan-copeland, line 43: typo maximium

PR #41, opened 2020-08-17
add transform function

kevinhkhuynh:
Added transformation function.

Takes in another function which can transform the vector anyway it wishes.

### DIFF
--- a/Vector.C
+++ b/Vector.C
@@ -134,20 +134,20 @@ Vector::operator = (const double& a)
 }
 
 Vector&
-Vector::transform(void (*f) (const int size, double* vector)) {
-      (*f)(d_dim, d_vec);
+Vector::transform(std::function<void(const int size, double* vector)> transformer) {
+      transformer(d_dim, d_vec);
       return *this;
 }
 
 void
-Vector::transform(Vector& result, void (*f) (const int size, double* vector)) const {
+Vector::transform(Vector& result, std::function<void(const int size, double* vector)> transformer) const {
       result.setSize(d_dim);
       result.setDistributed(d_distributed);
-      (*f)(d_dim, result.getVector());
+      transformer(d_dim, result.getVector());
 }
 
 void
-Vector::transform(Vector*& result, void (*f) (const int size, double* vector)) const {
+Vector::transform(Vector*& result, std::function<void(const int size, double* vector)> transformer) const {
       // If the result has not been allocated then do so.  Otherwise size it
       // correctly.
       if (result == 0) {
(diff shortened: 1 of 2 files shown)


kevinhkhuynh:
still in progress

kevinhkhuynh:
Hi, everyone. No need to do a serious look, but I would appreciate feedback on the function interfaces, and if you would like any additional functions, as you will be the primary users of these functions. Thank you.

kevinhkhuynh:
I've verified that this works on a basic operator I created in Laghos.

dylan-copeland:
Thanks Kevin, but instead of `void (*f) (const int size, double* vector)` I had in mind a function of the form `void (*f) (const int size, double* vectorIn, double* vectorOut)`. Maybe it does not make much difference, because the external code (e.g. Laghos) would need to have a special function written for this purpose, which could be of either form. For example, in MFEM operators typically have the form `Mult(const mfem::Vector &x, mfem::Vector &y)`, and a new version would need to be written of the form `Mult(double* x, double* y)`, which could alternatively be of the form `Mult(const int size, double* x)` with an auxiliary vector `y` used within and then copied to x. For convenience, it may better to support both versions (one vector and two vector arguments).

kevinhkhuynh:
> Thanks Kevin, but instead of `void (*f) (const int size, double* vector)` I had in mind a function of the form `void (*f) (const int size, double* vectorIn, double* vectorOut)`. Maybe it does not make much difference, because the external code (e.g. Laghos) would need to have a special function written for this purpose, which could be of either form. For example, in MFEM operators typically have the form `Mult(const mfem::Vector &x, mfem::Vector &y)`, and a new version would need to be written of the form `Mult(double* x, double* y)`, which could alternatively be of the form `Mult(const int size, double* x)` with an auxiliary vector `y` used within and then copied to x. For convenience, it may be better to support both versions (one vector and two vector arguments).

No problem, Dylan. It is done. There is a slight concern that the class is littered with too many transform functions now, but I think the added convenience may be worth it.

goxberry:
Use `std::function` instead of a function pointer; the former admits lambdas, functions, classes with appropriate definitions of `operator()`, or anything satisfying the `Callable` concept. The latter only works with functions.

kevinhkhuynh:
That is a very good idea to make the transformer function more general, Geoffrey. I've gone ahead and switched to std::function.

chldkdtn:
> Use `std::function` instead of a function pointer; the former admits lambdas, functions, classes with appropriate definitions of `operator()`, or anything satisfying the `Callable` concept. The latter only works with functions.

Glad to see that you are contributing, Geoff!

PR #44, opened 2020-09-14
Basis generator options

kevinhkhuynh:
Put StaticSVDBasisGenerator and IncrementalSVDBasisGenerator options into a struct. I debated adding a constructor to the struct options, but if we did, the backwards compatibility problems would still exist once we added new parameters and Laghos/Ardra still used the old constructor. So, the user will have to put in the options into a struct one-by-one, similar to how it is done in Laghos.

### DIFF
--- a/IncrementalSVD.h
+++ b/IncrementalSVD.h
@@ -81,7 +81,6 @@ struct IncrementalSVDOptions
    sampling_tol(sampling_tol_),
    max_time_between_samples(max_time_between_samples_) {}
 
-
    int dim;
    double linearity_tol;
    int max_basis_dimension;
--- a/IncrementalSVD.h
+++ b/IncrementalSVD.h
@@ -19,7 +19,7 @@
 
 namespace CAROM {
 
-struct IncrementalSVDOptions
+struct IncrementalSVDOptions : virtual public SVDOptions
 {
   /**
    * @brief Constructor.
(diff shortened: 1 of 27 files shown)


IncrementalSVDBasisGenerator.C
```diff
@@ -18,43 +18,28 @@
...
-   d_svdsampler.reset(new IncrementalSVDSampler(dim,
-                                                linearity_tol,
-                                                skip_linearly_dependent,
-                                                fast_update,
-                                                max_basis_dimension,
-                                                initial_dt,
-                                                samples_per_time_interval,
-                                                sampling_tol,
-                                                max_time_between_samples,
+   d_svdsampler.reset(new IncrementalSVDSampler(options.dim,
```
dylan-copeland, line 26: There may be many classes that have many constructor arguments, and `IncrementalSVDSampler` is just one, but what do you think about simplifying its constructor as well by passing `options`? If `IncrementalSVDSampler` is an internal class not called by users typically, then it is not so bad, but the libROM code would still be cleaner with a struct for all these arguments.

kevinhkhuynh: Okay, I see. Yes, you're right. I'll do the same thing for both incremental and static and just pass down the options struct, picking out what is needed for the class.

kevinhkhuynh: done

dylan-copeland:
@kevinhkhuynh The code looks good, and I like how it eliminates hundreds of lines of code. Have you done some testing to verify nothing is broken? Does libROM have regression tests, or does the rom-dev branch of Laghos need changes to use this new version of libROM so that the Laghos regression tests can be used to verify the libROM changes?

kevinhkhuynh:
> 
> 
> @kevinhkhuynh The code looks good, and I like how it eliminates hundreds of lines of code. Have you done some testing to verify nothing is broken? Does libROM have regression tests, or does the rom-dev branch of Laghos need changes to use this new version of libROM so that the Laghos regression tests can be used to verify the libROM changes?

libROM has basic tests (which are passing), but not to the depth of regression tests. I can create a new branch in Laghos using these options and this branch of libROM, and run the regression tests there against rom-dev using the master branch of libROM

kevinhkhuynh:
All Laghos regression tests pass.

IncrementalSVD.h
```diff
@@ -19,6 +19,87 @@
...
+   IncrementalSVDOptions(int dim_, double linearity_tol_, int max_basis_dimension_, double initial_dt_, int samples_per_time_interval_, double sampling_tol_, double max_time_between_samples_) :
+   dim(dim_),
+   linearity_tol(linearity_tol_),
+   max_basis_dimension(max_basis_dimension_),
+   initial_dt(initial_dt_),
+   samples_per_time_interval(samples_per_time_interval_),
+   sampling_tol(sampling_tol_),
+   max_time_between_samples(max_time_between_samples_) {}
+
+
```
dylan-copeland, line 84: Extra line

dylan-copeland (approved):

kevinhkhuynh:
all laghos tests still pass.

I realized while trying to update Laghos and Ardra that it can be tedious to construct the SVDOptions struct, especially if a user is only creating one basis generator, so the new change allow users to also construct the SVDBasisGenerator in a similar way as they did before. However, they can still also do it like in Laghos, changing the options one-by-one. When we add new parameters, we should add it at the end of the parameter list of the SVDOptions constructor and give it a default value, and backwards compatibility will be retained until the API is more finalized.

chldkdtn (approved):
Looks great! Thank you Kevin.

PR #45, opened 2020-09-28
Laghos compile mac

kevinhkhuynh:
Update the laghos_compile script to automatically install MAC packages as well, so that we can easily install dependencies in Laghos by doing "make setup".

And also fix a few bugs from the max_time_intervals pull request.


chldkdtn (approved):
Looks good to me.

PR #46, opened 2020-10-12
Bug/carom verify

kevinhkhuynh:
1. Fix small cout bug relating to svtol
2. Change CAROM_ASSERT to CAROM_VERIFY for Matrix, Vector, SVD, and BasisGenerator classes.
3. Leave CAROM_ASSERT for any access functions, such as Matrix.item() and Vector.item()

### DIFF
--- a/IncrementalSVDFastUpdate.C
+++ b/IncrementalSVDFastUpdate.C
@@ -164,7 +164,9 @@ IncrementalSVDFastUpdate::computeBasis()
        std::cout << "d_num_rows_of_W = " << d_num_rows_of_W << "\n";
        std::cout << "d_singular_value_tol = " << d_singular_value_tol << "\n";
        std::cout << "smallest SV = " << d_S->item(d_num_samples-1,d_num_samples-1) << "\n";
-       std::cout << "next smallest SV = " << d_S->item(d_num_samples-2,d_num_samples-2) << "\n";
+       if (d_num_samples > 1) {
+            std::cout << "next smallest SV = " << d_S->item(d_num_samples-2,d_num_samples-2) << "\n";
+       }
    }
    // remove the smallest singular value if it is smaller than d_singular_value_tol
    if ( (d_singular_value_tol != 0.0) &&
--- a/IncrementalSVDFastUpdate.C
+++ b/IncrementalSVDFastUpdate.C
@@ -147,12 +147,6 @@ void
 IncrementalSVDFastUpdate::computeBasis()
 {
    d_basis = d_U->mult(d_Up);
-   /*
-   if (fabs(checkOrthogonality(d_basis)) >
-       std::numeric_limits<double>::epsilon()*static_cast<double>(d_num_samples)) {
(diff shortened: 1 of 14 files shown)


dylan-copeland (approved):

chldkdtn (approved):

IncrementalSVDFastUpdate.C
```diff
@@ -184,7 +186,7 @@ IncrementalSVDFastUpdate::computeBasis()
 
        if (d_updateRightSV)
        {
-           Matrix* d_basis_right_new = new Matrix(d_num_rows_of_W, d_num_samples-1, false);
+           Matrix* d_basis_right_new = new Matrix(d_num_rows_of_W, d_num_samples-1, d_basis_right->distributed());
```
kevinhkhuynh, line 189: @chldkdtn Hi Youngsoo. In this function, should we also remove the last row and column of d_S whenever a singular_value is below d_singular_value_tol?

kevinhkhuynh: Never mind. It seems like it's not needed.

chldkdtn:
@kevinhkhuynh sv_tol is basically implementing Lines 18 through 20 of Algorithm 2 in the space-time ROM paper, i.e., https://www.sciencedirect.com/science/article/pii/S0021999120306197 although we did not use that option effectively.

IncrementalSVDFastUpdate.C
```diff
@@ -341,22 +341,6 @@ IncrementalSVDFastUpdate::addNewSample(
       max_U_dim = d_total_dim;
    }
 
-   if (fabs(checkOrthogonality(d_U)) >
-       std::numeric_limits<double>::epsilon()*static_cast<double>(max_U_dim)) {
-      reOrthogonalize(d_U);
-   }
-   if (fabs(checkOrthogonality(d_Up)) >
```
kevinhkhuynh, line 346: Hi Youngsoo. Here, d_U and d_Up are orthogonalized which ensure d_basis is orthogonalized afterwards. So when svtol is turned off, d_basis remain orthogonal. But when svtol is turned on, if a column is removed, d_basis is modified. I'm not too familiar with orthogonality, but it seems d_basis may not be orthogonal after a column is removed.

chldkdtn:
@kevinhkhuynh looks good! Can you check if all the regression tests in ARDRA ROM pass with this fix?

kevinhkhuynh:
@chldkdtn Please confirm the additional

> 
> 
> @kevinhkhuynh looks good! Can you check if all the regression tests in ARDRA ROM pass with this fix?

The regression tests still pass. svtol was set too small such that it never activated in Ardra.

chldkdtn:
> @chldkdtn Please confirm the additional
> 
> > @kevinhkhuynh looks good! Can you check if all the regression tests in ARDRA ROM pass with this fix?
> 
> The regression tests still pass. svtol was set too small such that it never activated in Ardra.

You can merge when you think is appropriate.

PR #47, opened 2020-10-21
QDEIM with QR via scalapack

dylan-copeland:
QDEIM is implemented, as an alternative to DEIM and GNAT, with oversampling done by GappyPOD+E (see https://arxiv.org/abs/1808.10473). The first samples (numbering up to the basis dimension) for QDEIM are computed using parallel QR with column pivoting for the transposed basis matrix, via `pdgeqpf` in scalapack.

The only validation of this code is its application in Laghos, which has some good results but does not truly validate the QR or QDEIM implementations.

### DIFF
--- a/Matrix.C
+++ b/Matrix.C
@@ -973,7 +973,7 @@ const
   
   initialize_matrix(&slpk, d_num_cols, row_offset[d_num_procs], 1, d_num_procs, 1, blocksize);  // transposed
 
-  CAROM_VERIFY(d_num_cols == pivots_requested); // Otherwise, should we just find the QR of the submatrix consisting of the first (pivots_requested) columns?
+  CAROM_VERIFY(d_num_cols <= pivots_requested); // Otherwise, should we just find the QR of the submatrix consisting of the first (pivots_requested) columns?
   
   for (int rank = 0; rank < d_num_procs; ++rank) {
     // Take the row-major data in d_mat and put it in a transposed column-major array in slpk
@@ -989,11 +989,17 @@ const
 
   // Just gather the pivots to root and discard the factorization
   CAROM_VERIFY(pivots_requested <= QRmgr.ipivSize);
-  CAROM_VERIFY(pivots_requested <= std::min(d_num_rows, d_num_cols));
+  CAROM_VERIFY(pivots_requested <= std::max(d_num_rows, d_num_cols));
 
   const int scount = std::max(0, std::min(pivots_requested, row_offset[my_rank+1]) - row_offset[my_rank]);
   int *mypivots = (scount > 0) ? new int[scount] : NULL;
 
+  { // Sanity check
+    int sumcount = 0;
+    MPI_Reduce(&scount, &sumcount, 1, MPI_INT, MPI_SUM, 0, MPI_COMM_WORLD);
+    CAROM_VERIFY(sumcount == pivots_requested || my_rank > 0);
+  }
+  
   for (int i=0; i<scount; ++i)
     mypivots[i] = QRmgr.ipiv[i]-1;  // make it 0-based
   
--- a/Matrix.C
+++ b/Matrix.C
@@ -337,8 +337,9 @@ Matrix::pointwise_mult(
    const Vector& other,
    Vector& result) const
 {
-   CAROM_VERIFY(!result.distributed());
-   CAROM_VERIFY(!distributed());
+   // TODO: change the CAROM_ASSERTs to CAROM_VERIFYs or generalize and eliminate the checks.
+   CAROM_ASSERT(!result.distributed());
+   CAROM_ASSERT(!distributed());
    CAROM_VERIFY(!other.distributed());
    CAROM_VERIFY(numColumns() == other.dim());
 
--- a/Matrix.C
+++ b/Matrix.C
@@ -1044,7 +1044,7 @@ const
   else
     {
       for (int i=0; i<scount; ++i)
-	row_pivot[i] = QRmgr.ipiv[i]-1;  // make it 0-based. TODO: is this used on non-root processes?
+	row_pivot[i] = QRmgr.ipiv[i]-1;  // make it 0-based. TODO: is this used on non-root processes? Not in QDEIM!
     }
   
   free_matrix_data(&slpk);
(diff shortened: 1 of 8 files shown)


chldkdtn:
@dylan-copeland This is awesome! Is this ready for review?

dylan-copeland:
> @dylan-copeland This is awesome! Is this ready for review?

@chldkdtn Not ready yet. There will also be a small PR for Laghos to use QDEIM. We should start using PR labels (e.g. "Work-in-progress", "Ready-for-review").

Matrix.C
```diff
@@ -335,9 +337,9 @@ Matrix::pointwise_mult(
    const Vector& other,
    Vector& result) const
 {
-   CAROM_ASSERT(!result.distributed());
-   CAROM_ASSERT(!distributed());
-   CAROM_ASSERT(!other.distributed());
+   CAROM_VERIFY(!result.distributed());
+   CAROM_VERIFY(!distributed());
```
kevinhkhuynh, line 341: In Ardra, there is an example of a distributed temporal basis matrix being pointwise multiplied with a non-distributed vector, and so setting this to VERIFY breaks the code. What do we think about only setting dimension checks to VERIFY (which is already done in the other merged PR) and leaving distributed checks as ASSERT?

dylan-copeland: If the condition is `false`, with either `ASSERT` or `VERIFY`, then something is wrong. It sounds like the check itself is wrong, if it works in a case where the condition is `false`. That is, if a distributed matrix times a non-distributed vector works, then we should just remove the check entirely. In this case, if in Ardra you have a distributed matrix, and one of its rows is multiplied by a non-distributed vector, I'm not sure how that could work. The row index is local, so how is it specified in parallel? This function is supposed to multiply the vector by the row, which is stored entirely on the process owning that row, but that process would be the only one having the correct result and would need to do `MPI_Bcast` to broadcast the result to all processes. I'm guessing there is a peculiar usage that makes this work, but it doesn't sound right in general. Can you provide more details?

kevinhkhuynh: @chldkdtn Perhaps Youngsoo can provide more accurate information on the meaning of the Ardra code and whether it can be rewritten to not break the ASSERT/VERIFY. 

Currently, the code is working, and for space-time ROM, there is an non-distributed initialized vector called carom_red_soln_init. It seems this vector can be distributed, but right now, it is not. The full order solution is projected onto the reduced order vector via: projectLiftedSolution(carom_red_soln_init, carom_soln, spatialbasis);. Then the reduced order solution is scaled by the temporal basis via: scaleWithTemporalBasis(carom_red_soln_init, temporalbasis, 0), where only row = 0 is chosen. This scaling function uses pointwise_mult and throws an error due to temporalbasis failing !distributed().

dylan-copeland: So local row 0 on every process scales the same undistributed vector on every process, with different results on every process? Is it used on every process?

kevinhkhuynh: Hi Dylan. You were right. It seems that temporal basis is distributed in name, but actually not distributed as all the processors have the same temporal basis. However, the spatial basis is distributed. And so, since they were created from one basis generator, both were set to distributed. I don't think libROM currently allows to set rightSV to not distributed while leftSV is.

chldkdtn: Looks like both of you have figured out what is going on with temporal basis. It is currently duplicated over the processors. If we can generalize the case, e.g., allowing distributed spatial basis and non-distributed temporal basis, that would be wonderful.

dylan-copeland: It sounds like the temporal basis should be non-distributed, in which case no changes to the `Matrix` class would be needed (just changes in the basis generator). I am not yet familiar with this issue or a test that could verify a libROM change is working, so I propose just keeping this as an `ASSERT` in this PR and letting @kevinhkhuynh fix this for Ardra in another PR. Does that sound good @kevinhkhuynh @chldkdtn ?

chldkdtn: @dylan-copeland @kevinhkhuynh Please remind me again. ASSERT becomes only active with debug mode, while VERIFY is active both with optimized and debug modes, right?

kevinhkhuynh: @dylan-copeland That sounds good to me.

@chldkdtn ASSERT is only active in debug mode, while VERIFY is active in all modes.

QDEIM.C
```diff
@@ -15,46 +15,418 @@
...
-  // processor that owns each sampled row, and fills f_basis_sampled with the
+  // processor that owns each sampled row, and fills f_basis_sampled_inv with the
   // sampled rows of the basis of the RHS.
 
-  // Compute the number of basis vectors used; this number can't
-  // exceed the number of columns of the matrix.
-  int num_basis_vectors_used = std::min(num_f_basis_vectors_used,
-					f_basis->numColumns());
+  CAROM_VERIFY(num_f_basis_vectors_used == f_basis->numColumns());  // The QR implementation uses the entire matrix.
+  CAROM_VERIFY(f_basis->numColumns() <= num_samples_req && num_samples_req <= f_basis->numRows());
```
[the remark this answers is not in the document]
chldkdtn: > I have a similar line in GNAT that is failing due to num_samples possibly being greater than f_basis->numRows for parallel runs since numRows() only returns the number of rows on the current processor. I've created a numDistributedRows function in the tests/gnatTest PR. Perhaps once we merge that PR, we can use that function here too.

@kevinhkhuynh Just to make sure, numDistributedRows return global number of rows, right?

siuwuncheung (approved):

dylan-copeland:
@chldkdtn @kevinhkhuynh reminder that this is the only QDEIM PR waiting for your review.

QDEIM.C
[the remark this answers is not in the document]
kevinhkhuynh: @dylan-copeland I can make the change to this line in my test_qdeim branch once test_gnat is finished next week, so this PR isn't held up. How does that sound?

kevinhkhuynh (approved):

QDEIM.C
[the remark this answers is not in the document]
chldkdtn: That sounds good.

chldkdtn (approved):

PR #52, opened 2020-12-30
Feature/randomized svd

kevinhkhuynh:
Implementation of randomized SVD using scalapack and qrcp.


chldkdtn:
@kevinhkhuynh This is great! I hope that you will include some tests for randomized SVD.

kevinhkhuynh:
> 
> 
> @kevinhkhuynh This is great! I hope that you will include some tests for randomized SVD.

Yes. Some tests will be included in this PR.

kevinhkhuynh:
@chldkdtn @dylan-copeland @siuwuncheung This took longer than I thought, but it's ready for review now. Uses pdeqpf to obtain Q in elementary householder format, than uses pdormqr to compute the action of Q. Originally, I computed the action in libROM, but it was way too slow.

Test results were computed using the Matlab script given by Youngsoo. libROM matches with these 2 small tests for the left basis, right basis, and singular values.

Matrix.C
```diff
@@ -704,7 +740,7 @@ Matrix::print(const char * prefix)
 {
    int my_rank;
    const bool success = MPI_Comm_rank(MPI_COMM_WORLD, &my_rank);
-   CAROM_VERIFY(success);
+   CAROM_ASSERT(success);
```
dylan-copeland, line 743: What is the issue with `CAROM_VERIFY` here?

Matrix.h
```diff
@@ -702,6 +705,37 @@ class Matrix
       void
       inverse();
 
+      /**
+       * @brief Returns a column of the matrix.
+       *
+       * @return A column of the matrix.
+       */
```
dylan-copeland, line 712: Maybe comment that the returned vector needs to be deleted by the user (not owned by `Matrix`).

RandomizedSVD.C
```diff
@@ -0,0 +1,267 @@
...
+
+     // Project snapshot matrix onto random subspace
+     Matrix* rand_proj = snapshot_matrix->mult(rand_mat);
+
+     // Get QR factorization of random projection
+     int *row_offset = new int[d_num_procs + 1];
+     row_offset[d_num_procs] = rand_proj->numDistributedRows();
+
+     int my_rank;
+     MPI_Comm_rank(MPI_COMM_WORLD, &my_rank);
```
dylan-copeland, line 109: Isn't this the same as `d_rank`?

RandomizedSVD.C
```diff
@@ -0,0 +1,267 @@
...
+
+     SLPK_Matrix svd_input;
+     make_similar_matrix(&svd_input, snapshot_matrix_transposed->numDistributedRows(),
+            snapshot_matrix_transposed->numColumns(), QRmgr.A->ctxt, d_blocksize, d_blocksize);
+
+     for (int rank = 0; rank < d_num_procs; ++rank) {
+        scatter_block(&svd_input, 1, row_offset[rank] + 1,
+                  snapshot_matrix_transposed->getData(), row_offset[rank + 1] - row_offset[rank],
+                  snapshot_matrix_transposed->numColumns(), rank);
+
```
dylan-copeland, line 163: Extra line

tests/test_RandomizedSVD.C
```diff
@@ -0,0 +1,176 @@
+
+
+/******************************************************************************
+ *
+ * Copyright (c) 2013-2019, Lawrence Livermore National Security, LLC
```
dylan-copeland, line 5: Should we update to 2021?

tests/test_RandomizedSVD.C
```diff
@@ -0,0 +1,176 @@
...
+  for (int i = 0; i < 9; i++) {
+    EXPECT_EQ(d_basis_vals[i], basis_true_ans[i]);
+  }
+
+  for (int i = 0; i < 9; i++) {
+    EXPECT_EQ(sv_vals[i], sv_true_ans[i]);
+  }
+}
+
+int main(int argc, char* argv[])
```
dylan-copeland, line 163: Nice testing! Would it be hard to add a parallel randomized SVD test? Are there any parallel SVD tests yet?

RandomizedSVD.C
```diff
@@ -0,0 +1,267 @@
...
+                  1, rand_proj->getData(), row_offset[rank + 1] - row_offset[rank],
+                  rand_proj->numColumns(), rank);
+     }
+
+     QRManager QRmgr;
+     qr_init(&QRmgr, &slpk_rand_proj);
+     qrfactorize(&QRmgr);
+
+     // Manipulate lq_A to get elementary household reflectors.
+     Matrix* lq_A = new Matrix(QRmgr.A->mdata, d_subspace_dim,
```
dylan-copeland, line 145: Is this used, or is a scalapack routine used instead?

RandomizedSVD.C
```diff
@@ -0,0 +1,267 @@
...
+     delete_factorizer();
+
+     int num_rows = d_total_dim;
+     int num_cols = d_num_samples;
+     if (d_subspace_dim < 1 || d_subspace_dim > std::min(num_rows, num_cols)) {
+       d_subspace_dim = std::min(num_rows, num_cols);
+     }
+
+     // Get snapshot matrix in distributed format.
+     // If there are less dimensions than samples, use the transpose instead.
```
dylan-copeland, line 56: Is there testing of both cases?

RandomizedSVD.C
```diff
@@ -0,0 +1,267 @@
...
+    double*, int*, double*, double*, int*, double*, int*, int*);
+}
+
+namespace CAROM {
+
+RandomizedSVD::RandomizedSVD(
+   Options options) :
+   StaticSVD(options),
+   d_subspace_dim(options.randomized_subspace_dim) {
+     srand(1);
```
dylan-copeland, line 40: It may be useful to allow the user to set the seed, so that different random numbers can be generated on each run.

dylan-copeland:
Thanks @kevinhkhuynh, this is a great new addition to libROM. There is significant new memory management, so it would be a good idea to run some small tests (serial and parallel) with valgrind to verify there are no memory leaks.

Matrix.C
kevinhkhuynh: For some reason, the verify was failing even though the rank was successfully retrieved, so I just changed it to an assert.

tests/test_RandomizedSVD.C
kevinhkhuynh: There are no parallel SVD tests yet, and this would be useful. I'll look into figuring out how to do parallel tests with Gtest, and add it to this PR.

RandomizedSVD.C
kevinhkhuynh: pdormqr assumes that the input Q contains the elementary householder reflections. Since the output to pdegpf combines Q and R into one matrix, we need to remove R by zeroing out some values and changing some values to 1, which is done in the following lines.

RandomizedSVD.C
kevinhkhuynh: Yes, currently there are 4 serial tests. One for full k, and one for smaller k for both cases (rows > cols and cols > rows)

dylan-copeland (approved):

tests/test_RandomizedSVD.C
chldkdtn: Yes! definitely.

tests/test_RandomizedSVD.C
chldkdtn: parallel SVD tests sounds great!

chldkdtn:
@kevinhkhuynh Can you double check if travis actually uses gtest and all our gtest suites are tested in travis?

kevinhkhuynh:
The ability to run serial tests with multiple processes have been added and the results are identical to the serial tests.

> 
> 
> @kevinhkhuynh Can you double check if travis actually uses gtest and all our gtest suites are tested in travis?

Yes. If you look here at one of the Travis tests run and scroll to the bottom (https://travis-ci.com/github/LLNL/libROM/jobs/468930047), you can see the gtest tests running and passing.

chldkdtn:
@kevinhkhuynh How did you check if parallel randomized SVD is correct? Did you compare the solution with serial randomized SVD run?

kevinhkhuynh:
>  How did you check if parallel randomized SVD is correct? Did you compare the solution with serial randomized SVD run?

Yes, the parallel tests run the serial tests with multiple processes and checks that the solution is the same with the serial randomized SVD run. There is an additional check to make sure that the basis is the correct size on each processor (each processor's basis should be smaller than the original, and since this passes we know it isn't running in serial).

kevinhkhuynh:
> 
> 
> > How did you check if parallel randomized SVD is correct? Did you compare the solution with serial randomized SVD run?
> 
> Yes, the parallel tests run the serial tests with multiple processes and checks that the solution is the same with the serial randomized SVD run. There is an additional check to make sure that the basis is the correct size on each processor (each processor's basis should be smaller than the original, and since this passes we know it isn't running in serial).

But I still need to run more tests before I change the label to RFR.

chldkdtn:
> > How did you check if parallel randomized SVD is correct? Did you compare the solution with serial randomized SVD run?
> 
> Yes, the parallel tests run the serial tests with multiple processes and checks that the solution is the same with the serial randomized SVD run. There is an additional check to make sure that the basis is the correct size on each processor (each processor's basis should be smaller than the original, and since this passes we know it isn't running in serial).

For both parallel and serial run, I think you need to use the same random matrix in order to get identical (or close to identical) results. Do you use the same random matrix?

kevinhkhuynh:
> 
> 
> > > How did you check if parallel randomized SVD is correct? Did you compare the solution with serial randomized SVD run?
> > 
> > 
> > Yes, the parallel tests run the serial tests with multiple processes and checks that the solution is the same with the serial randomized SVD run. There is an additional check to make sure that the basis is the correct size on each processor (each processor's basis should be smaller than the original, and since this passes we know it isn't running in serial).
> 
> For both parallel and serial run, I think you need to use the same random matrix in order to get identical (or close to identical) results. Do you use the same random matrix?

Yes, randomizedSVD has a parameter that sets a random seed, and the random seed is defaulted to 1 unless the user changes it.

chldkdtn (approved):

siuwuncheung (approved):

PR #57, opened 2021-03-04
Feature/greedy algorithm

kevinhkhuynh:
Implementing the greedy algorithm.

1. User constructs the greedy algorithm class, given it a domain of parameter points. (A set of CAROM::Vectors or doubles).
2. User requests a parameter point to sample. To keep things general, libROM does not return the Vector/Double corresponding to the parameter point, but an index to the parameter point set. The user should already have access to this since they supplied it when constructing the greedy algorithm class. If not, they can obtain it by calling the get method in the libROM class.
3. In another code base, the user creates a ROM for this parameter point.
4. User then asks greedy algorithm which point it should find a residual for. This is how the greedy algorithm decides where to create the next ROM. libROM gives that to the user as an index into the parameter point set.
5. In another code base, the user computes the residual and gives it to libROM.
6. Steps 4 and 5 repeat, until libROM returns a -1 for the next point. Meaning, we have decided on a new point to create a ROM at or we have not yet converged with a random subset of parameter points meeting the error threshold. We need to go back to step 1.
7. This process repeats until libROM returns a -1 for the next point to sample at step 1. The algorithm has converged and our ROM database is complete.

The greedy algorithm class does not maintain any basis for any ROMs. The user does this in their own code, since the greedy algorithm class does not need this information. The user only needs it to calculate the residual in step 5.

The user interface for this is a little complex since this does not really seem to be an algorithm, but more of a workflow. Any suggestions to simplify it are appreciated.


kevinhkhuynh:
Ready for review.

GreedyParameterPointSelector.h
```diff
@@ -0,0 +1,286 @@
...
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ *
+ *****************************************************************************/
+
+// Description: This class greedily selects parameter points
+//              for the construction of a ROM database. The implemented
+//              greedy algorithm is obtained from algorithm 2 of Choi et. al's
+//              paper "Gradient-based constrained optimization using a database
+//              of linear reduced-order models".
```
dylan-copeland, line 15: Should we add the arxiv link?

GreedyParameterPointSelector.h
```diff
@@ -0,0 +1,286 @@
...
+#include <algorithm>
+
+namespace CAROM {
+
+class BasisGenerator;
+class Matrix;
+class Vector;
+
+/**
+ * Class GreedyParameterPointSelector is an class defining the interface a
```
dylan-copeland, line 47: Two typos in this line

GreedyParameterPointSelector.h
```diff
@@ -0,0 +1,286 @@
...
+       */
+      std::vector<Vector> d_parameter_points;
+
+      /**
+       * @brief The parameter points that were already sampled.
+       */
+      std::set<int> d_parameter_sampled_indices;
+
+      /**
+       * @brief The parameter point indices (used to generate the random subsets)
```
dylan-copeland, line 178: Not a big deal, but there is inconsistent use of periods in these comments.

GreedyParameterPointSelector.h
```diff
@@ -0,0 +1,286 @@
...
+      * @brief The size of the subset of parameter points used per iteration.
+      */
+      int d_subset_size;
+
+     /**
+      * @brief The size of the subset of parameter points used to check convergence.
+      */
+      int d_convergence_subset_size;
+
+
```
dylan-copeland, line 212: empty line

GreedyParameterPointSelector.h
```diff
@@ -0,0 +1,286 @@
...
+      */
+      int d_next_point_requiring_residual;
+
+     /**
+      * @brief Whether the use the centroid heuristic for obtaining the first
+      *        parameter point.
+      */
+      bool d_use_centroid;
+
+      /**
```
dylan-copeland, line 229: Indenting increases here. It doesn't need to be fixed. I just want to point out that having `make style` in libROM would be nice.

GreedyParameterPointSelector.C
```diff
@@ -0,0 +1,465 @@
...
+    int random_seed,
+    bool debug_algorithm)
+{
+    //convert parameter_points from double to Vector
+    CAROM_VERIFY(param_space_max > param_space_min);
+    CAROM_VERIFY(param_space_size > 0);
+
+    std::vector<Vector> parameter_points_vec;
+    Vector vec(1, false);
+    if (param_space_size == 1)
```
dylan-copeland, line 80: Why are the values 1 and 2 for `param_space_size` treated specially? It will not be clear to the user, and should at least be documented. It would be simpler just to use the else case only, with a positive integer value for `param_space_size`

GreedyParameterPointSelector.C
```diff
@@ -0,0 +1,465 @@
...
+        vec.item(0) = parameter_points[i];
+        parameter_points_vec.push_back(vec);
+    }
+
+    constructObject(parameter_points_vec, tolerance, saturation,
+        subset_size, convergence_subset_size, use_centroid, random_seed, debug_algorithm);
+}
+
+GreedyParameterPointSelector::GreedyParameterPointSelector(
+    double param_space_min,
```
dylan-copeland, line 63: Thanks for adding this constructor. We could more generally construct a box of dimension `d` by taking a vector of min, max, and size of length `d`. For now, this should be fine, and if there is common usage of higher dimensional boxes we can add that constructor later.

GreedyParameterPointSelector.C
```diff
@@ -0,0 +1,465 @@
...
+        // obtain the centroid and find the point closest to centroid
+        if (d_use_centroid)
+        {
+            Vector centroid(d_parameter_points[0]);
+            for (int i = 1; i < d_parameter_points.size(); i++) {
+                centroid += d_parameter_points[i];
+            }
+            centroid.mult(1.0 / d_parameter_points.size(), centroid);
+
+            double min_dist_to_centroid = INT_MAX;
```
dylan-copeland, line 177: This should work in all practical cases, but it would be a bug if the minimum distance is greater than `INT_MAX` (obviously should not happen). It would be safer to initialize this to the distance for `i==0`.

GreedyParameterPointSelector.C
```diff
@@ -0,0 +1,465 @@
...
+    return d_next_point_requiring_residual;
+}
+
+void
+GreedyParameterPointSelector::setPointResidual(double error, int rank, int num_procs)
+{
+    CAROM_VERIFY(error >= 0);
+    CAROM_VERIFY(d_point_requiring_residual_computed);
+
+    double *proc_errors = new double[num_procs];
```
dylan-copeland, line 304: Instead of allocating this array and gathering, you can just do `MPI_Allreduce` on the scalar local error squared.

GreedyParameterPointSelector.C
kevinhkhuynh: There's a strange case where if a user gives a range, but only with a size of 1, using the else will return just the bottom value. Similary for 2, it will return the bottom and top value, and nothing in the middle. Maybe there's a better segmenting equation that can do the right thing for all cases.

dylan-copeland: Yes, those are some tricky cases. I suppose we should try to provide the simplest interface for the user to understand. I'm not sure what the user would have in mind when specifying 1. If they really want 1 point, they could set the min and max to be the same. In the case of 2 points, the min and max values make sense (the else case). Just using the else case makes sense to me in general, provided it is documented.

GreedyParameterPointSelector.h
kevinhkhuynh: I do plan to add a 'make style' using astyle, but I think it would be better when there are no open PR's as I tried it once before and it ended up with a bunch of complicated merge conflicts, so I reverted.

GreedyParameterPointSelector.C
dylan-copeland: What you implemented so far was to use the specified number of points in the interior (excluding endpoints) for the cases 1 and 2, but including the endpoints for cases 3 and higher. Maybe a `bool` to specify whether to include endpoints would be more general and clearer. Documenting the case of "endpoints=true" and "size=1 or 2" would also be helpful.

chldkdtn:
@siuwuncheung @dylan-copeland Let's review and approve this soon.

GreedyParameterPointSampler.h
```diff
@@ -0,0 +1,734 @@
...
+//              for the construction of a ROM database. The implemented
+//              greedy algorithm is derived from algorithm 2 of Choi et. al's
+//              paper "Gradient-based constrained optimization using a database
+//              of linear reduced-order models": https://arxiv.org/abs/1506.07849
+//
+//              The greedy algorithm workflow is as follows:
+//              1. Construct the GreedyParameterPointSampler by giving it a
+//                 domain of parameter points.
+//              2. Request a parameter point to sample.
+//              3. Request a parameter point to compute an error error indicator for.
```
dylan-copeland, line 21: typo

GreedyParameterPointSampler.h
```diff
@@ -0,0 +1,734 @@
...
+// Description: This class greedily selects parameter points
+//              for the construction of a ROM database. The implemented
+//              greedy algorithm is derived from algorithm 2 of Choi et. al's
+//              paper "Gradient-based constrained optimization using a database
+//              of linear reduced-order models": https://arxiv.org/abs/1506.07849
+//
+//              The greedy algorithm workflow is as follows:
+//              1. Construct the GreedyParameterPointSampler by giving it a
+//                 domain of parameter points.
+//              2. Request a parameter point to sample.
```
dylan-copeland, line 20: It is not clear what it means to sample a parameter point. Some more details would help.

GreedyParameterPointSampler.h
```diff
@@ -0,0 +1,734 @@
...
+        std::string output_log_path = "",
+        std::string warm_start_file_name = "",
+        bool use_centroid = true,
+        int random_seed = 1,
+        bool debug_algorithm = false);
+
+    /**
+     * @brief Constructor.
+     *
+     * @param[in] param_space_min A CAROM::Vectors representing the minimum
```
dylan-copeland, line 135: `A CAROM::Vector`

GreedyParameterPointSampler.h
```diff
@@ -0,0 +1,734 @@
...
+ *       a ROM database efficiently.
+ */
+class GreedyParameterPointSampler
+{
+public:
+    /**
+     * @brief Constructor.
+     *
+     * @param[in] parameter_points A vector of CAROM::Vectors containing
+     *                             the different parameter points.
```
dylan-copeland, line 68: Would it be more precise to say, "A vector of CAROM::Vectors containing many user-defined parameter points from which a subset will be chosen for building ROMs"?

GreedyParameterPointSampler.h
```diff
@@ -0,0 +1,734 @@
...
+     *
+     * @return A struct holding the point and the location of the nearest local
+     *         ROM or NULL if no point is required.
+     */
+    struct GreedyErrorIndicatorPoint getNextConvergencePointRequiringErrorIndicator();
+
+    /**
+     * @brief Generate a vector of random points.
+     *
+     * @param[in] num_points The number of opints to generate
```
dylan-copeland, line 409: typo

GreedyParameterPointSampler.C
```diff
@@ -0,0 +1,1420 @@
...
+int
+getNearestPointIndex(std::vector<Vector> param_points, Vector point)
+{
+
+    double closest_dist_to_points = INT_MAX;
+    int closest_point_index = -1;
+
+    for (int i = 0; i < param_points.size(); i++)
+    {
+        Vector diff;
```
dylan-copeland, line 59: If there are many points, this could be optimized by some geometric search algorithm (e.g. maintaining a data structure that decomposes the domain into boxes and keeps the points in those boxes), instead of this exhaustive O(n) search. However, the greatest cost seems to be associated with evaluating ROMs at each parameter point? Or are ROMs only evaluated at a small subset of points? It is not clear to me yet whether such an optimization of this search would help in the end, considering other costs of the overall algorithm. In any case, it is not necessary in this PR, but something to think about maybe for the future.

GreedyParameterPointSampler.C
```diff
@@ -0,0 +1,1420 @@
...
+            d_convergence_points.push_back(point);
+        }
+    }
+    database.close();
+
+    rng.seed(d_random_seed + d_parameter_sampled_indices.size());
+}
+
+void
+GreedyParameterPointSampler::constructParameterPoints()
```
dylan-copeland, line 394: This function just checks some conditions and prints out information, whereas `constructParameterPoints` in the derived classes actually constructs points. It seems better to rename this as something about checking input rather than constructing points. Also, that would avoid having the same name for different functions (not virtual).

GreedyParameterPointSampler.C
```diff
@@ -0,0 +1,1420 @@
...
+    }
+    if (d_counter == d_convergence_subset_size)
+    {
+        return createGreedyErrorIndicatorPoint(nullptr, nullptr);
+    }
+
+    d_next_point_requiring_error_indicator = -1;
+
+    //get next point requiring error indicator
+    while (d_counter < (int) d_convergence_points.size())
```
dylan-copeland, line 801: `while` could be `if`?

GreedyParameterPointSampler.C
```diff
@@ -0,0 +1,1420 @@
...
+        {
+            point.item(j) = unif[j](rng);
+        }
+        random_points.push_back(point);
+    }
+    return random_points;
+}
+
+std::shared_ptr<Vector>
+GreedyParameterPointSampler::getNearestROM(Vector point)
```
dylan-copeland, line 1198: It looks like some code duplication could be avoided by calling `getNearestROMIndex` from this function.

GreedyParameterPointSampler.h
```diff
@@ -0,0 +1,734 @@
...
+     *                  iteration.
+     * @param[in] max_clamp A scalar factor representing the maximum amount
+     *                      the error indicator tolerance can change per iteration.
+     * @param[in] subset_size The size of the random subset.
+     * @param[in] convergence_subset_size The size of the convergence subset.
+     * @param[in] use_centroid Whether to use the centroid heuristic when
+     *                         determining the first parameter point to sample.
+     * @param[in] random_seed A random seed.
+     * @param[in] debug_algorithm Whether to turn off all randomness for
+     *                            debugging purposes.
```
dylan-copeland, line 82: Doesn't `random_seed` determine the randomness, so that `debug_algorithm` is unnecessary? I see that `debug_algorithm` determines whether a shuffle is done, but that uses `rng` which depends on `random_seed`.

GreedyParameterPointSampler.h
```diff
@@ -0,0 +1,734 @@
...
+public:
+    /**
+     * @brief Constructor.
+     *
+     * @param[in] parameter_points A vector of CAROM::Vectors containing
+     *                             the different parameter points.
+     * @param[in] check_local_rom Compute local ROM error indicator each iteration.
+     * @param[in] relative_error_tolerance The relative error tolerance value
+     *                                     for which to end the algorithm.
+     * @param[in] alpha A alpha constant to increase greedy algorithm by each
```
dylan-copeland, line 72: Would it be more accurate to say "A constant factor to increase the error indicator tolerance"?

tests/test_GreedyParameterPointPreDefinedSampler.C
```diff
@@ -0,0 +1,144 @@
...
+    int mpi_init, d_rank, d_num_procs;
+    MPI_Initialized(&mpi_init);
+    if (mpi_init == 0) {
+        MPI_Init(nullptr, nullptr);
+    }
+
+    MPI_Comm_rank(MPI_COMM_WORLD, &d_rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &d_num_procs);
+
+    std::vector<double> paramPoints = {1.0, 2.0, 3.0, 99.0, 100.0, 101.0};
```
dylan-copeland, line 42: Is there any testing of multidimensional parameters, in libROM or Laghos?

GreedyParameterPointSampler.h
kevinhkhuynh: You're right. With random_seed, debug_algorithm is unnecessary.

tests/test_GreedyParameterPointPreDefinedSampler.C
kevinhkhuynh: added test here, simple for just 2 dimensions.

GreedyParameterPointSampler.h
kevinhkhuynh: Actually, I had to put it back. It seems without it, linux and mac are untestable since they may shuffle differently even with the random seed.

GreedyParameterPointSampler.C
kevinhkhuynh: They're slightly different to the point where it would be difficult to combine them, so I changed the function name and comment description in .h instead to better match what it is doing.

GreedyParameterPointSampler.C
dylan-copeland: Sounds good. By the way, I think the idea I am referring to is called "spatial hashing."

dylan-copeland (approved):

siuwuncheung (approved):

PR #64, opened 2021-05-26
Feature/introduce mfem

kevinhkhuynh:
New instructions to compile are in the README.md.


kevinhkhuynh:
It seems we will need to add mesh files too. Perhaps the person who creates the MFEM/libROM example can add that in their PR.

kevinhkhuynh:
To test the MFEM examples. Can cd into build and run ./ex1p and ./ex16p after building libROM with the -m option.

dylan-copeland (approved):
Thanks Kevin, this looks good. One comment I have, which does not need to be addressed in this PR, is that `SampleMesh.hpp` has implementation specific for two spaces (L2, H1), which is valid for mixed diffusion (L2 for pressure, H1 for velocity) and hydrodynamics (L2 for energy, H1 for x, v), but not completely general. For ex1p, we may need a version for a single space (H1). That is something we can do in a future PR.

kevinhkhuynh:
> 
> 
> Thanks Kevin, this looks good. One comment I have, which does not need to be addressed in this PR, is that `SampleMesh.hpp` has implementation specific for two spaces (L2, H1), which is valid for mixed diffusion (L2 for pressure, H1 for velocity) and hydrodynamics (L2 for energy, H1 for x, v), but not completely general. For ex1p, we may need a version for a single space (H1). That is something we can do in a future PR.

Sounds good to me.

chldkdtn (approved):

chldkdtn:
@kevinhkhuynh @dylan-copeland @siuwuncheung Let's say if any PR got at least two approvals, then the PR can be merged. How is that?

siuwuncheung (approved):

chldkdtn:
@kevinhkhuynh This needs to be merged.

PR #69, opened 2021-07-21
DMD 

kevinhkhuynh:
DMD. Works for serial and parallel.


kevinhkhuynh:
Currently, only works in serial. To try it out, build and run:

./ex10
 
It is defaulted to ef 0.99, but you can change -ef 0.XX and -rdim X. If both ef and rdim are set, ef will be ignored and rdim will be used.

Currently, all the example does is train and predict the last time step using DMD, outputting the relative error.

kevinhkhuynh:
Works in serial and parallel. To try it out, build and run:

srun -n X -p pdebug ex10 -no-vis

It is defaulted to ef 0.99, but you can use -ef 0.XX and -rdim X. If both ef and rdim are set, ef will be ignored and rdim will be used.

Currently, all the example does is train and predict the last time step using DMD, outputting the relative error.

kevinhkhuynh:
The files for interest to review are DMD.h, DMD.C, and ex10p.cpp

kevinhkhuynh:
Self-reminder to push in changes from feature/dmd2

chldkdtn (approved):

chldkdtn:
@siuwuncheung @dylan-copeland This is pretty cool. Let's merge to the master soon. Please review and approve in your earliest time.

DMD.h
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * Copyright (c) 2013-2021, Lawrence Livermore National Security, LLC
+ * and other libROM project developers. See the top-level COPYRIGHT
+ * file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ *
+ *****************************************************************************/
+
+// Description:Computes the DMD algorithm on the given snapshot matrix.
```
dylan-copeland, line 11: Space after `:`

DMD.h
```diff
@@ -0,0 +1,177 @@
...
+class Vector;
+
+class DMD
+{
+public:
+
+    /**
+     * @brief Constructor.
+     */
+    DMD(int dim);
```
dylan-copeland, line 31: `dim` should be documented as full-order state dimension.

DMD.h
```diff
@@ -0,0 +1,177 @@
...
+     */
+    int d_rank;
+
+    /**
+     * @brief The number of processors being run on.
+     */
+    int d_num_procs;
+
+    /**
+     * @brief The total dimensions of the sample vector.
```
dylan-copeland, line 117: "total dimensions" -> dimension

DMD.h
```diff
@@ -0,0 +1,177 @@
...
+    DMD(int dim);
+
+    /**
+     * @brief Sample the new state, u_in.
+     *
+     * @pre u_in != 0
+     *
+     * @param[in] u_in The state at the specified time.
+     *
+     * @return True if the sampling was successful.
```
dylan-copeland, line 40: What does `false` return value of `takeSample` mean, and how should it be handled? This should be documented.

DMD.h
```diff
@@ -0,0 +1,177 @@
...
+    void projectInitialCondition(const Vector* init);
+
+    /**
+     * @brief Predict state given a time. Uses the projected initial condition of the
+     *        training dataset (the first column).
+     *
+     * @param[in] t The time of the outputted state.
+     * @param[in] dt The delta time of the simulation.
+     */
+    Vector* predict(double t, double dt);
```
dylan-copeland, line 68: My understanding is that `n = t/dt` is used, not `t` or `dt` individually. Would it make more sense to have one input for `n` (which could be a `double`), representing the number of online timesteps? Isn't `dt` associated with an offline simulation and assumed to be the same for online? Variable dt is not allowed, right? Or could variable dt work somehow?

DMD.C
```diff
@@ -0,0 +1,489 @@
...
+ *
+ * Copyright (c) 2013-2021, Lawrence Livermore National Security, LLC
+ * and other libROM project developers. See the top-level COPYRIGHT
+ * file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ *
+ *****************************************************************************/
+
+// Description: Interface to the DMD algorithm.
```
dylan-copeland, line 11: `Interface to` -> `Implemention of`?

DMD.C
```diff
@@ -0,0 +1,489 @@
...
+
+    delete f_snapshots;
+}
+
+void
+DMD::constructDMD(const Matrix* f_snapshots,
+                  int d_rank,
+                  int d_num_procs)
+{
+    Matrix* f_snapshots_minus = new Matrix(f_snapshots->numRows(),
```
dylan-copeland, line 101: It would be better (and necessary for large problems) to avoid making these two copies of the snapshot matrix, but I realize this may be too much work for this PR. It seems having the copies makes the code simpler, but it is more expensive in terms of memory. This may be fine for the initial implementation, but it should at least be documented that this should be improved in the future for large problems. The amount of work to avoid the copies may be significant. In particular the "minus" snapshot matrix is passed to scalapack, so we would need to extend the scalapack interface to work on a submatrix. Also, we would need `CAROM::Matrix` multiplication for submatrices. It doesn't have to be done in this PR, but some comments would be helpful.

DMD.C
```diff
@@ -0,0 +1,489 @@
...
+                                           f_snapshots->numColumns() - 1, f_snapshots->distributed());
+    Matrix* f_snapshots_plus = new Matrix(f_snapshots->numRows(),
+                                          f_snapshots->numColumns() - 1, f_snapshots->distributed());
+
+    // Break up snapshots into snapshots_minus and snapshots_plus
+    // snapshots_minus = all columns of snapshots except last
+    // snapshots_plus = all columns of snapshots except first
+    for (int i = 0; i < f_snapshots->numRows(); i++)
+    {
+        f_snapshots_minus->item(i, 0) = f_snapshots->item(i, 0);
```
dylan-copeland, line 111: This loop could be simplified (just use +1 in "plus" matrix, with `j` starting at 0).

DMD.C
```diff
@@ -0,0 +1,489 @@
...
+    // into column major order.
+    for (int row = 0; row < d_k; ++row) {
+        for (int col = row+1; col < d_k; ++col) {
+            double tmp = A_tilde->item(row, col);
+            A_tilde->item(row, col) = A_tilde->item(col, row);
+            A_tilde->item(col, row) = tmp;
+        }
+    }
+
+    // Now call lapack to do the eigensolve.
```
dylan-copeland, line 243: Would it be better to refactor the eigensolve into a new function that could be reused?

DMD.C
```diff
@@ -0,0 +1,489 @@
...
+    Vector* d_projected_init_real_2 = d_phi_imaginary_squared->mult(rhs_imaginary);
+    d_projected_init_real = d_projected_init_real_1->minus(d_projected_init_real_2);
+
+    Vector* d_projected_init_imaginary_1 = d_phi_real_squared->mult(rhs_imaginary);
+    Vector* d_projected_init_imaginary_2 = d_phi_imaginary_squared->mult(rhs_real);
+    d_projected_init_imaginary =  d_projected_init_imaginary_1->plus(d_projected_init_imaginary_2);
+
+    delete d_phi_real_squared;
+    delete d_phi_real_squared_2;
+    delete  d_projected_init_real_1;
```
dylan-copeland, line 390: Two spaces after `delete`

DMD.C
```diff
@@ -0,0 +1,489 @@
...
+
+    // This block does the actual ScaLAPACK call to do the factorization.
+    svd_init(d_factorizer.get(), &svd_input);
+    d_factorizer->dov = 1;
+    factorize(d_factorizer.get());
+    free_matrix_data(&svd_input);
+
+    // Compute how many basis vectors we will actually use.
+    int num_singular_vectors = std::min(f_snapshots_minus->numColumns(), f_snapshots_minus->numDistributedRows());
+    if (d_energy_fraction != -1)
```
dylan-copeland, line 166: This must work, but it still worries me a little to compare a double and int. Why not make it `-1.0`?

DMD.C
```diff
@@ -0,0 +1,489 @@
...
+    CAROM_VERIFY(d_snapshots[0].dim() > 0);
+    for (int i = 0 ; i < d_snapshots.size() - 1; i++)
+    {
+        CAROM_VERIFY(d_snapshots[i].dim() == d_snapshots[i + 1].dim());
+        CAROM_VERIFY(d_snapshots[i].distributed() == d_snapshots[i + 1].distributed());
+    }
+
+    Matrix* snapshot_mat = new Matrix(d_snapshots[0].dim(), d_snapshots.size(), d_snapshots[0].distributed());
+
+    // This might be slow since we're accessing a different memory address each time.
```
dylan-copeland, line 477: Is it clear from testing whether this is a bottleneck, or would timing need to be done for some large problems?

kevinhkhuynh: Good timing. I just ran into an issue with Youngsoo with slow DMD. This is not a bottleneck. It is quite fast. The bottleneck is the scalapack SVD call.

examples/dg_advection.cpp
```diff
@@ -0,0 +1,988 @@
+//                       libROM MFEM Example: DG Advection
```
dylan-copeland, line 1: Since the MFEM examples were given new names, it is not clear which MFEM example these examples correspond to. A comment like "Adapted from MFEM example ex??.cpp" would be helpful.

examples/dg_advection.cpp
```diff
@@ -0,0 +1,988 @@
...
+
+    // 15. Calculate the relative error between the DMD final solution and the true solution.
+    Vector dmd_solution_u(result_u->getData(), result_u->dim());
+    Vector true_solution_u(U->GetData(), U->Size());
+    Vector diff_u(true_solution_u.Size());
+    subtract(dmd_solution_u, true_solution_u, diff_u);
+
+    double* diff_norm_u = new double[num_procs] {};
+    double proc_diff_norm_u = diff_u.Norml2();
+    MPI_Gather(&proc_diff_norm_u, 1, MPI_DOUBLE, diff_norm_u, 1, MPI_DOUBLE, 0, MPI_COMM_WORLD);
```
dylan-copeland, line 720: Calculating global `l2` norms from local norms in this way is terribly inconvenient, but fortunately there is a function in MFEM to do this. You can use `sqrt(InnerProduct(MPI_COMM_WORLD, diff_u, diff_u));` to simplify all these examples.

examples/dg_advection.cpp
```diff
@@ -0,0 +1,988 @@
...
+        double tot_true_solution_u_norm = 0;
+        for (int i = 0; i < num_procs; i++)
+        {
+            tot_diff_norm_u += std::pow(diff_norm_u[i], 2);
+            tot_true_solution_u_norm += std::pow(true_solution_u_norm[i], 2);
+        }
+        tot_diff_norm_u = std::sqrt(tot_diff_norm_u);
+        tot_true_solution_u_norm = std::sqrt(tot_true_solution_u_norm);
+
+        std::cout << "Relative error of solution (u) at t_final: " << t_final << " is " << tot_diff_norm_u / tot_true_solution_u_norm << std::endl;
```
dylan-copeland, line 737: "Relative error of DMD solution"
would be clearer (also in other files)

examples/nonlinear_elasticity.cpp
```diff
@@ -0,0 +1,868 @@
+//                       libROM MFEM Example: Nonlinear elasticity
+//
+// Compile with: make nonlinear_elasticity
+//
+// Sample runs:
+//    mpirun -np 4 nonlinear_elasticity -s 3 -rs 2 -dt 3
```
dylan-copeland, line 6: The error in this test is over 100%
srun -n 1 -p pdebug nonlinear_elasticity -s 3 -rs 2 -dt 3
I would not expect all the sample runs to have good results with DMD, but at least one sample run for each example should work. Did you find a case that works?

chldkdtn: I am actually working on this, @copeland. I will post the working examples today.

chldkdtn: Try the following:

./dg_advection -p 3 -rp 1 -dt 0.005 -tf 4 -visit
mpirun -n 8 ./dg_euler -p 2 -rs 2 -rp 1 -o 1 -s 3 -visit
mpirun -np 8 ./heat_conduction -s 3 -a 0.5 -k 0.5 -o 4 -tf 0.7 -vs 1 -visit
mpirun -np 8 ./nonlinear_elasticity -s 2 -rs 1 -dt 0.01 -tf 5 -visit

dylan-copeland: Yes, these work. I wonder why some cases do not work for DMD, but we don't have to answer that now. There should be documentation on sample runs that work for DMD.

DMD.h
kevinhkhuynh: You're right. I think this would work better.

docs/html/class_c_a_r_o_m_1_1_d_m_d-members.html
```diff
@@ -0,0 +1,113 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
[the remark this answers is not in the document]
dylan-copeland: What is the decision on this? When I build, git shows lots of modified html and js files. Should these be in the repo?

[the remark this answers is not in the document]
kevinhkhuynh: The html and js files are needed, but perhaps they don't need to be pushed as often. Let me see if there is a way to only have it show up when the content of the webpage has actually changed rather than unimportant metadata.

examples/dg_advection.cpp
kevinhkhuynh: Thanks. I wish I knew this two weeks ago!

DMD.C
kevinhkhuynh: Added TODO.

examples/nonlinear_elasticity.cpp
kevinhkhuynh: There was a small bug with dg_advection. The results are a lot better now after my latest commit. It should work well with the default "srun -n 8 -p pdebug dg_advection". The other examples should be unaffected.

kevinhkhuynh: Added example run documentation to the top of each file.

dylan-copeland (approved):
Thanks @kevinhkhuynh, this is really nice work, and the results are very good.

siuwuncheung (approved):

PR #73, opened 2021-08-20
Feature/greedy algorithm example

kevinhkhuynh:
Add global serial greedy algorithm example (poisson_greedy)


chldkdtn (approved):
Very nice!

PR #77, opened 2021-09-14
Sample mesh for arbitrary number of spaces

dylan-copeland:
The sample mesh construction is generalized to take into account an arbitrary number of finite element spaces and their samples. The DOF maps between these spaces and the corresponding sample mesh spaces are also constructed. The function for gathering sampled rows of a distributed matrix is also updated for these changes, and it is changed to operate on one matrix in one space at a time, to avoid redundant work.

The API is necessarily changed.

There is a plan to change the API further by making a class for the functions in this file, which would reduce the arguments and be more user-friendly. That will be a big change, postponed to a future PR.


dylan-copeland:
Reminder that this PR and Laghos PR 162 are ready for review. @chldkdtn @kevinhkhuynh @siuwuncheung

kevinhkhuynh:
Would it be possible to add doxygen documentation to this file or will you do that in the next PR?

dylan-copeland:
> Would it be possible to add doxygen documentation to this file or will you do that in the next PR?

Very good point. There definitely should be some good documentation, since even I barely understand this code. It would be better to wait until the next PR when this code should be completely refactored, so that the documentation would be completely different.

kevinhkhuynh:
> 
> 
> > Would it be possible to add doxygen documentation to this file or will you do that in the next PR?
> 
> Very good point. There definitely should be some good documentation, since even I barely understand this code. It would be better to wait until the next PR when this code should be completely refactored, so that the documentation would be completely different.

Sounds good to me.

kevinhkhuynh (approved):

kevinhkhuynh:
Would you mind updating the Changelog in the main directory? We should probably start using that.

PR #79, opened 2021-10-15
Feature/matrix interpolation

kevinhkhuynh:
1. Change the directory structure of the examples and tests.
2. Update the CMakeLists.txt to work with spack.
3. Update any missing documentation.
4. Implement matrix interpolation.
5. Use matrix implementation in the dg_advection example.
6. Add a global_rom dg_advection example.
7. Separate out dmd, global_rom, local_rom examples if they were all in the same example file.


Vector.C
```diff
@@ -532,4 +532,60 @@ Vector::local_read(const std::string& base_file_name, int rank)
     database.close();
 }
 
+int getCenterPoint(std::vector<Vector*> points,
```
dylan-copeland, line 535: Can this function replace the similar code in `GreedyParameterPointSampler::getNextParameterPoint()`?

DMD.h
```diff
@@ -21,6 +21,9 @@ namespace CAROM {
 class Matrix;
 class Vector;
 
+/**
+ * Class DMD implements Brunton's DMD algorithm on a given snapshot matrix.
```
dylan-copeland, line 25: Is there a paper that could be referenced here?

Interpolater.h
```diff
@@ -0,0 +1,132 @@
...
+
+#include <vector>
+
+namespace CAROM {
+
+class Matrix;
+class Vector;
+
+/**
+ * Interpolater is a uninstantiable protected class that retains common
```
dylan-copeland, line 24: a -> an

Interpolater.h
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * Copyright (c) 2013-2021, Lawrence Livermore National Security, LLC
+ * and other libROM project developers. See the top-level COPYRIGHT
+ * file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ *
+ *****************************************************************************/
+
+// Description: Computes the Interpolater algorithm on the given snapshot matrix.
```
dylan-copeland, line 11: Interpolater or interpolator? According to the dictionary, both are acceptable English words, but it seems interpolator is commonly used in mathematics. A google search for interpolator shows many results, but a google search for interpolater asks "Did you mean interpolator?" There are many instances to change, but if you agree that interpolator is the more common choice then it would be better to make the change now rather than later.

Interpolater.C
```diff
@@ -0,0 +1,181 @@
...
+
+std::vector<double> Interpolater::obtainRBF(Vector* point)
+{
+    // Inverse quadratic RBF
+    std::vector<double> inv_q;
+    for (int i = 0; i < d_parameter_points.size(); i++)
+    {
+        Vector diff;
+        point->minus(*d_parameter_points[i], diff);
+        double res = 1.0 / (1.0 + std::pow(d_epsilon*diff.norm(), 2));
```
dylan-copeland, line 65: I'm not sure if this function is a bottleneck, but `x*x` may be significantly faster than `pow(x, 2)`.

dylan-copeland: Even better would be to implement a `norm2` function for the norm squared, which avoids the even more costly `sqrt` call and the `pow` call.

Interpolater.C
```diff
@@ -0,0 +1,181 @@
...
+    MPI_Comm_size(MPI_COMM_WORLD, &num_procs);
+
+    std::vector<Matrix*> rotation_matrices;
+
+    // Obtain the rotation matrices to rotate the bases into
+    // the same generalized coordinate space.
+    for (int i = 0; i < parameter_points.size(); i++)
+    {
+        CAROM_VERIFY(bases[i]->numRows() == bases[ref_point]->numRows());
+        CAROM_VERIFY(bases[i]->numColumns() == bases[ref_point]->numColumns());
```
dylan-copeland, line 95: What about the case that different parameters have different basis dimensions?

kevinhkhuynh: I'm not sure how that would work since we need to multiply the bases together, and also need to multiply the reduced matrices together, so they would have to be the same dimensions

Interpolater.C
```diff
@@ -0,0 +1,181 @@
...
+        {
+            Matrix* identity_matrix = new Matrix(bases[i]->numColumns(), bases[i]->numColumns(), false);
+            for (int j = 0; j < identity_matrix->numColumns(); j++) {
+                identity_matrix->item(j, j) = 1.0;
+            }
+            rotation_matrices.push_back(identity_matrix);
+            continue;
+        }
+
+        Matrix* basis_mult_basis = bases[i]->transposeMult(bases[ref_point]);
```
dylan-copeland, line 110: Isn't this always a non-distributed, small matrix? If so, is the SVD computation really done in parallel?

Interpolater.h
kevinhkhuynh: Oh, I didn't know there were two different spellings lol. I don't have a preference, so I'll switch to interpolator.

Interpolater.C
kevinhkhuynh: Yes, transposeMult always returns a undistributed matrix, so the SVD computation is not done in parallel.

Interpolater.C
chldkdtn: yep. they need to be the same dimension.

dylan-copeland: The dimension determined by an energy fraction usually varies for different parameters. In that case, is it necessary to take the maximum of all dimensions, in order to use this interpolation method?

Interpolater.C
dylan-copeland: The basis matrices are large and distributed row-wise, so the transposeMult is performed in parallel and produces a small matrix of size equal to the number of basis vectors. This is fine, but my question was about whether the SVD should be done in parallel by scalapack. If it is small, then lapack can do the SVD in serial. In libROM we use scalapack for SVD of a large distributed matrix (e.g. `StaticSVD`), not for small matrices like this.

Interpolater.C
chldkdtn: Yep, we have to use the maximum of all dimensions. Instead of energy fraction criteria, there is an option for specifying a specific number for the reduced space dimension, right?

Interpolater.C
chldkdtn: Is serial SVD with scalapack is particularly slower than lapack?

VectorInterpolator.C
```diff
@@ -0,0 +1,164 @@
...
+        // Perform log mapping by doing Y - X.
+        std::vector<Vector*> gammas;
+
+        for (int i = 0; i < d_parameter_points.size(); i++)
+        {
+            // For ref point, gamma is the zero vector.
+            if (i == d_ref_point)
+            {
+                Vector* gamma = new Vector(d_rotated_reduced_vectors[d_ref_point]->dim(), d_rotated_reduced_vectors[d_ref_point]->distributed());
+                gammas.push_back(gamma);
```
dylan-copeland, line 138: Is `gamma = 0.0;` necessary?

MatrixInterpolator.C
```diff
@@ -0,0 +1,408 @@
...
+    Matrix* exp_eigs = new Matrix(exp_eigenpair.eigs.size(), exp_eigenpair.eigs.size(), false);
+    for (int i = 0; i < exp_eigenpair.eigs.size(); i++)
+    {
+        exp_eigs->item(i, i) = std::exp(exp_eigenpair.eigs[i]);
+    }
+
+    exp_ev->inverse(exp_ev_inv);
+    Matrix* exp_ev_mult_exp_eig = exp_ev->mult(exp_eigs);
+
+    // Expoentiate gamma
```
dylan-copeland, line 258: typo

MatrixInterpolator.C
```diff
@@ -0,0 +1,408 @@
...
+        // Perform log mapping.
+        std::vector<Matrix*> gammas;
+
+        for (int i = 0; i < d_parameter_points.size(); i++)
+        {
+            // For ref_point, gamma is the zero matrix
+            if (i == d_ref_point)
+            {
+                Matrix* gamma = new Matrix(ref_matrix_inv->numRows(), ref_matrix_inv->numColumns(), ref_matrix_inv->distributed());
+                gammas.push_back(gamma);
```
dylan-copeland, line 292: `gamma = 0.0;`?

MatrixInterpolator.C
```diff
@@ -0,0 +1,408 @@
...
+            }
+            else
+            {
+                Matrix* y_mult_ref_matrix_inv = d_rotated_reduced_matrices[i]->mult(ref_matrix_inv);
+
+                // Diagonalize YX^-1 to obtain log of this matrix.
+                EigenPair log_eigenpair = SymmetricRightEigenSolve(y_mult_ref_matrix_inv);
```
dylan-copeland, line 299: Could you please document how `log` of a `Matrix` is defined (theoretically) and computed in practice? It seems `log` and `exp` could be member functions of `Matrix` (non-distributed case), rather than just blocks of code here in the interpolator.

MatrixInterpolator.h
```diff
@@ -0,0 +1,141 @@
...
+     * @brief Constructor.
+     *
+     * @param[in] parameter_points The parameter points.
+     * @param[in] rotation_matrices The rotation matrices associated with
+     *                              each parameter point.
+     * @param[in] reduced_matrices The reduced matrices associated with
+     *                             each parameter point.
+     * @param[in] ref_point The index within the vector of parameter points
+     *                      to the reference point
+     * @param[in] matrix_type The type of matrix (R = real, B = basis,
```
dylan-copeland, line 46: It's not clear what `R` and `B` are. There are 3 methods of interpolation, `NS`, `SPD`, and general (Y-X), so I don't see how these 4 matrix types correspond to the 3 methods. More documentation on the matrix and interpolation types would be helpful.

Interpolater.C
dylan-copeland: Not sure about performance, but at least the code for lapack seems simpler and shorter (compare QDEIM.C and Interpolator.C). Maybe the SVD can be put into a function to be more compact and avoid repetition.

Interpolater.C
kevinhkhuynh: Yes, that is better. I'll change to do that. Previously, I was ensuring the basis dimensions had the same reduced dimension beforehand.

MatrixInterpolator.C
kevinhkhuynh: log, exp, and i think square root and many others are calculated the same way. to get the log of a matrix M, diagonalize M, then do

eigenvectors * log/exp/sqrt/etc.(eigenvalues) * inverse(eigenvectors).

so yes, that part can be moved into Matrix. I'll see about how to do that.

MatrixInterpolator.h
kevinhkhuynh: R = general (Y-X)
B = basis. For basis, doing the rotation is a bit different, then with general Y-X. I'll add more documentation

MatrixInterpolator.C
kevinhkhuynh: The constructor of a new matrix by default sets every value to 0.0.

VectorInterpolator.C
kevinhkhuynh: I talked to Youngsoo about this. It seems it is required. Perhaps he can expand. @chldkdtn

MatrixInterpolator.C
dylan-copeland: I checked the constructors for `Matrix`, and it looks like it just calls `setSize`, which allocates memory without initializing. Am I missing something?

Interpolater.C
kevinhkhuynh: We need to specify the reduced space dimension and not use energy fraction criteria (which is what I currently instruct the user to do when running the problem). Ignore my previous reply.

MatrixInterpolator.C
kevinhkhuynh: I think we talked about this in another PR. Currently, setSize destroys the original data and creates a new empty memory.

d_mat = new double [new_size] {};

The {} actually initializes everything to zero.

MatrixInterpolator.C
dylan-copeland: I'm not sure what conditions ensure that the eigenvalues of `YX^{-1}` are positive, so that log can be taken. There seems to be an assumption that two matrices `X` and `Y` are close enough in some sense that `YX^{-1}` is close to identity, with positive eigenvalues. This seems likely to fail in some cases, so an error message would be helpful, rather than just the `CAROM_VERIFY` below.

MatrixInterpolator.C
kevinhkhuynh: https://stackoverflow.com/questions/201101/how-to-initialize-all-members-of-an-array-to-the-same-value

kevinhkhuynh: To be safe, I can change {} to {0}

VectorInterpolator.C
dylan-copeland: I agree that `gamma` should be zero, but my question is whether the constructor initializes it to zero (as in the discussion about the `Matrix` constructor.

MatrixInterpolator.C
dylan-copeland: Thanks for clarifying, I did forget about this.

Matrix.h
```diff
@@ -137,7 +137,7 @@ class Matrix
 
     /**
      * @brief Sets the number of rows and columns of the matrix and
-     * reallocates storage if needed.
+     * reallocates storage if needed. All values are initalized to zero.
```
dylan-copeland, line 140: initalized -> initialized

MatrixInterpolator.C
```diff
@@ -0,0 +1,429 @@
...
+
+    // Obtain B matrix by calculating inverse quadratic distance.
+    Matrix* B = new Matrix(gammas.size(), gammas.size(), false);
+    for (int i = 0; i < B->numRows(); i++)
+    {
+        for (int j = i + 1; j < B->numColumns(); j++)
+        {
+            Vector diff;
+            d_parameter_points[i]->minus(*d_parameter_points[j], diff);
+            double res = 1.0 / (1.0 + std::pow(diff.norm(), 2));
```
dylan-copeland, line 118: norm2

MatrixInterpolator.C
```diff
@@ -0,0 +1,429 @@
...
+    Matrix* exp_eigs = new Matrix(exp_eigenpair.eigs.size(), exp_eigenpair.eigs.size(), false);
+    for (int i = 0; i < exp_eigenpair.eigs.size(); i++)
+    {
+        exp_eigs->item(i, i) = std::exp(exp_eigenpair.eigs[i]);
+    }
+
+    exp_ev->inverse(exp_ev_inv);
+    Matrix* exp_ev_mult_exp_eig = exp_ev->mult(exp_eigs);
+
+    // Exponetiate gamma
```
dylan-copeland, line 269: Exponentiate

VectorInterpolator.C
```diff
@@ -0,0 +1,164 @@
...
+
+    // Obtain B vector by calculating inverse quadratic distance.
+    Matrix* B = new Matrix(gammas.size(), gammas.size(), false);
+    for (int i = 0; i < B->numRows(); i++)
+    {
+        for (int j = i + 1; j < B->numColumns(); j++)
+        {
+            Vector diff;
+            d_parameter_points[i]->minus(*d_parameter_points[j], diff);
+            double res = 1.0 / (1.0 + std::pow(diff.norm(), 2));
```
dylan-copeland, line 90: norm2

MatrixInterpolator.h
```diff
@@ -0,0 +1,143 @@
...
+#include <string>
+
+namespace CAROM {
+
+class Matrix;
+class Vector;
+
+/**
+ * MatrixInterpolator interpolates reduced matrices of a set of parameter points
+ * and returns an interpolated reduced matrix for an unseen parameter point.
```
dylan-copeland, line 27: It is good that this general comment mentions that `MatrixInterpolator` is intended for reduced matrices only, but I would also recommend stating that performance depends on the matrix sizes being small, as the computational complexity is high, due to direct matrix inversion.

Vector.C
kevinhkhuynh: Good catch

examples/prom/mixed_nonlinear_diffusion.cpp
```diff
@@ -794,8 +794,8 @@ int main(int argc, char *argv[])
         vector<int> sample_dofs(nldim);  // Indices of the sampled rows
         CAROM::DEIM(FR_librom,
                     nldim,
-                    &sample_dofs[0],
-                    &num_sample_dofs_per_proc[0],
+                    sample_dofs,
```
kevinhkhuynh, line 797: @dylan-copeland I had to make some changes in this commit to the mixed nonlinear diffusion example to get the master branch to build. Does this look good to you?

dylan-copeland: Yes, thanks for updating.

kevinhkhuynh:
Reminder to self to add stuff Youngsoo mentioned

dylan-copeland (approved):
Thanks @kevinhkhuynh for this great work! I did not run the Laghos regression tests, but assuming the latest version of this PR passes the tests, I think it is ready.

kevinhkhuynh:
All Laghos tests pass.

chldkdtn (approved):

chldkdtn:
@siuwuncheung when you get a chance, would you review this PR?

siuwuncheung (approved):

PR #83, opened 2021-11-12
more build updates

kevinhkhuynh:
1. Switch to v1.0.0. Then will make release on libROM
2. Add small fix for mac installation
3. Add mesh files. They are very small, so I do not think there is much harm in using these instead and makes it easier if mfem is installed elsewhere and the examples need a hardcoded path to the meshes.
4. Add README.md in examples directory.


examples/README.md
```diff
@@ -0,0 +1,19 @@
+# Examples
+
+These data and examples are built using MFEM and derived from their examples
+(https://mfem.org/examples/). libROM's examples are divided into two categories.
+
+## DMD
```
dylan-copeland, line 6: Since this is an introductory README, it would be helpful to write out what DMD and PROM stand for, e.g.
`Dynamic mode decomposition (DMD)`

dylan-copeland (approved):

PR #97, opened 2022-02-18
Time windowing DMD

siuwuncheung:
Time windowing DMD for generic CSV data

### DIFF
--- a/lib/algo/DMD.C
+++ b/lib/algo/DMD.C
@@ -438,14 +438,14 @@ DMD::createSnapshotMatrix(std::vector<Vector*> snapshots)
 }
 
 void
-DMD::summary(std::string output_path)
+DMD::summary(std::string output_path, int window)
 {
     if (d_rank == 0)
     {
         CSVDatabase* csv_db(new CSVDatabase);
 
-        csv_db->putDoubleVector(output_path + "/singular_value.csv", d_sv, d_num_singular_vectors);
-        csv_db->putComplexVector(output_path + "/eigenvalue.csv", d_eigs, d_eigs.size());
+        csv_db->putDoubleVector(output_path + "/window" + std::to_string(window) + "_singular_value.csv", d_sv, d_num_singular_vectors);
+        csv_db->putComplexVector(output_path + "/window" + std::to_string(window) + "_eigenvalue.csv", d_eigs, d_eigs.size());
 
         delete csv_db;
     }
--- a/lib/algo/DMD.C
+++ b/lib/algo/DMD.C
@@ -438,14 +438,14 @@ DMD::createSnapshotMatrix(std::vector<Vector*> snapshots)
 }
 
 void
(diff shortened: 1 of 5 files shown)


lib/utils/CSVDatabase.h
```diff
@@ -254,6 +254,23 @@ class CSVDatabase : public Database
...
+    /**
+     * @brief Reads a vector of doubles associated with the supplied key
+     * from the currently open CSV database file.
+     *
+     * @pre !key.empty()
+     *
+     * @param[in] key The key associated with the vector of values to be
+     *                read.
+     * @param[out] data The allocated vector of double values to be read.
+     * @param[in] append True if append to the vector, otherwise overwite.
```
dylan-copeland, line 266: overwite -> overwrite

lib/utils/CSVDatabase.h
```diff
@@ -271,6 +288,18 @@ class CSVDatabase : public Database
...
+     * @brief Count the number of lines of CSV database file.
+     *
+     * @pre !key.empty()
+     *
+     * @param[in] key The key associated with the vector of strings to be
+     *                read.
+     */
+    int
+    getLineCount(
+        const std::string& key);
```
dylan-copeland, line 301: I see that other functions also use the argument name `key`, but isn't it just the filename? Would it be clearer to replace `key` with `filename`?

lib/algo/DMD.C
```diff
@@ -438,14 +438,14 @@ DMD::createSnapshotMatrix(std::vector<Vector*> snapshots)
 }
 
 void
-DMD::summary(std::string output_path)
+DMD::summary(std::string output_path, int window)
```
kevinhkhuynh, line 441: Would it be better to put  "/window" + std::to_string(window) in the output_path and keep this function the same as it was before, since some people may only use regular DMD and not time-windowing DMD?

lib/utils/CSVDatabase.h
siuwuncheung: I actually had the same question before. `filename` is more clear. I will amend the variable name in both `CSVDatabase` and `HDFDatabase`.

lib/algo/DMD.C
siuwuncheung: Thanks for the suggestion, Kevin. I will change that.

lib/utils/CSVDatabase.h
siuwuncheung: I see. I will leave HDFDatabase as is. Just to note that the name of the argument `file_name` for the member functions in `CSVDatabase` will be different from `key` in `Database`.

lib/utils/CSVDatabase.h
siuwuncheung: Resolved. Thanks, Dylan!

dylan-copeland (approved):

chldkdtn (approved):

PR #98, opened 2022-02-24
Add file to aid post processing 

jtlau:
Hello,

I've been using this file (`combine_samples.C`) to load in multiple snapshots (in my case, from parametric simulations), subtract the mean, and compute the SVD over the snapshots.

This is similar but extended capability to `tests/load_samples.C`. 

I'm not sure the best location for this helper file / function, but it was simple to add it to `tests/`. 

I think it might be useful to anyone who needs to work with parametric simulations. Subtracting the mean is optional. There is more description and an example command in the file, included here:

Description: This is a file function that reads snapshot or basis files
		      and computes the SVD over the group. Also provides the option
		      to subtract the mean of the group before computing the SVD.

Example: `mpirun -n 2 ./combine_samples -b -m file1 file2 file3`


chldkdtn:
@dylan-copeland @kevinhkhuynh @siuwuncheung, it sounds like combine_samples.C does the similar thing to our offset capability. Let's discuss where is the best place to put this file and how to merge it to the existing architecture of libROM.

dylan-copeland:
Thank you @jtlau for this nice example of how to combine snapshot files. We have similar scripts in codes like Laghos, which we typically call "merge" scripts, that basically read snapshot files along with some offset vector and compute the SVD on the combined or merged snapshots. I don't think we have an example of how to do this in libROM, so adding this example seems helpful to other users. 

Regarding the "mean" vector, it appears to be a vector read from file, not necessarily the mean. The mean must be computed in a different code, so this example does not need to call it the mean. More generally, it can be called an "offset" vector.

tests/combine_samples.C
```diff
@@ -0,0 +1,186 @@
...
+    if (!subtract_mean) {
+        /*-- Compute SVD and save file --*/
+        if (rank==0) std::cout << "Computing SVD" << std::endl;
+        int rom_dim = static_basis_generator->getSpatialBasis()->numColumns();
+        if (rank==0) std::cout << "U ROM Dimension: " << rom_dim << std::endl;
+        static_basis_generator->endSamples();
+    }
+    else {
+        // Close file so it can be opened:
+        static_basis_generator = nullptr;
```
dylan-copeland, line 119: Instead of writing and reading the same data, can you just use `static_basis_generator->getSnapshotMatrix`?

jtlau: Yes! I will change that.

jtlau:
Thanks @dylan-copeland for the comments. I am including an option to subtract an arbitrary offset file currently. Do you know what format the offset might likely be in? An hdf5 snapshot file?

dylan-copeland:
> Thanks @dylan-copeland for the comments. I am including an option to subtract an arbitrary offset file currently. Do you know what format the offset might likely be in? An hdf5 snapshot file?

Yes, hdf5 can be assumed. The libROM `Vector` class has the `read` and `write` functions for HDF files, which we use in Laghos for example.

jtlau:
@dylan-copeland I have included the option to subtract an offset, using the `Vector` `read` function. How does that look?

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
+/******************************************************************************
+ *
+ * Copyright (c) 2013-2021, Lawrence Livermore National Security, LLC
```
dylan-copeland, line 3: 2022

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+ * and other libROM project developers. See the top-level COPYRIGHT
+ * file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ *
+ *****************************************************************************/
+
+// Description: This is a file function that reads snapshot or basis files
```
dylan-copeland, line 11: This function

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+    MPI_Comm_size(MPI_COMM_WORLD, &size);
+    int rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+	
+    std::vector<std::string> sample_names;
+    int snaps = 0; 
+    int dim   = 0;
+    bool subtract_mean = false;
+    bool subtract_offset = false;
+    std::string kind = "snapshot";
```
dylan-copeland, line 58: Not a big deal, but `kind` could be an `enum`.

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+	    CAROM::Matrix *snapshots = (CAROM::Matrix*) reader.getSnapshotMatrix(0);
+	    dim    = snapshots->numRows();
+ 	    snaps += snapshots->numColumns();
+        }
+	else {
+            CAROM::Matrix *basis = (CAROM::Matrix*) reader.getSpatialBasis(0);
+	    dim    = basis->numRows();
+	    snaps += basis->numColumns();
+	}
+    }
```
dylan-copeland, line 107: The verification of `dim > 0` could go here, along with setting another variable `dimFirst = dim` if `dimFirst` is not already set, otherwise verification that `dim == dimFirst`. This will ensure all files have the same dimensions, which could help users prevent or find mistakes.

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+    static_basis_generator.reset(new CAROM::BasisGenerator(
+                                     CAROM::Options(dim, snaps).setMaxBasisDimension(snaps), false,
+                                     generator_filename));
+
+    if (rank==0) std::cout << "Loading data from " << kind << std::endl;
+    for(const auto& sample_name: sample_names) {
+        static_basis_generator->loadSamples(sample_name, kind);
+    }
+
+    if (rank==0) std::cout << "Saving data uploaded as a snapshot" << std::endl;
```
dylan-copeland, line 123: as a snapshot matrix

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+    }
+    else {
+    	/*-- load data from hdf5 file to find the mean and subtract it --*/
+    	if (rank==0) std::cout << "Reading snapshots" << std::endl;
+    	CAROM::Matrix *snapshots = (CAROM::Matrix*) static_basis_generator->getSnapshotMatrix();
+        
+    	int num_rows = snapshots->numRows();
+    	int num_cols = snapshots->numColumns();
+        CAROM::Vector *offset = new CAROM::Vector(num_rows, true);
+    	//double* offset = new double[num_rows];
```
dylan-copeland, line 141: remove

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+
+    if (rank==0) std::cout << "Loading data from " << kind << std::endl;
+    for(const auto& sample_name: sample_names) {
+        static_basis_generator->loadSamples(sample_name, kind);
+    }
+
+    if (rank==0) std::cout << "Saving data uploaded as a snapshot" << std::endl;
+    static_basis_generator->writeSnapshot();
+	
+    if (!subtract_mean && !subtract_offset) {
```
dylan-copeland, line 126: `CAROM_VERIFY(!(subtract_mean && subtract_offset));`

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+    	    }
+            offset->mult(1.0 / (double)num_cols, offset);
+		    offset->write("mean");
+    	}
+        else if (subtract_offset) {
+            if (rank==0) std::cout << "Subtracting offset from file: " << offset_file << std::endl;
+            offset->read(offset_file);
+        }
+
+        /*-- Subtract mean from snapshot/bases matrix --*/
```
dylan-copeland, line 160: Replace mean with offset, for more generality? Or say "Subtract mean or offset"?

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+    	CAROM::Matrix* snaps_mean = new CAROM::Matrix(num_rows, num_cols, false);
```
dylan-copeland, line 161: Instead of allocating another matrix, can you just subtract the offset from `snapshots`? For large matrices, this could be a significant amount of memory that is doubled.

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+    	    CAROM::Options(dim, snaps).setMaxBasisDimension(snaps), false,
+    	    generator_filename+"_sub"));
+    	
+    	for (int col = 0; col < num_cols; col++) {
+    	    double* snap_cur = new double[num_rows];
+    	  
+    	    for (int row = 0; row < num_rows; row++) {
+    	        snap_cur[row] = snaps_mean->item(row,col);
+    	    } 
+    	    static_basis_generator2->takeSample(snap_cur, 0.0, false);
```
dylan-copeland, line 181: It is unfortunate that another generator is necessary. Again, it seems 2x the necessary memory is being used. I don't see a good way to avoid this, unless we add the capability to modify existing snapshots. That is, to just subtract the offset from the snapshots already taken. Maybe this can be done in a future PR.

dylan-copeland (changes requested):
Thanks @jtlau for the option to subtract an offset. It looks good, and I just have some minor suggestions.

chldkdtn:
@dylan-copeland @kevinhkhuynh @siuwuncheung @jtlau, Can we think of combining "merge" in Laghos and "combineSnapshot" in SU2 into one script and include the script in libROM so that other physics solver can take advantage as well?

tests/combine_samples.C
```diff
@@ -0,0 +1,200 @@
...
+        return 1;
+    }
+ 
+    /*-- Read dimension and count number of snapshots/bases --*/
+    if (rank==0) std::cout << "Opening files to read dimension and count number of snapshots/bases" << std::endl;
+    for (const auto& sample_name: sample_names) {
+        CAROM::BasisReader reader(sample_name);
+		
+        if (kind == "snapshot") {
+	    CAROM::Matrix *snapshots = (CAROM::Matrix*) reader.getSnapshotMatrix(0);
```
dylan-copeland, line 98: It looks like this reads the entire matrix into memory, although you only need to get the dimensions here. The function `getSnapshotMatrix` first reads the dimensions and then reads the matrix. You could write a new function, `getSnapshotDimensions`, which just reads the dimensions. To avoid code duplication, `getSnapshotDimensions` could be called by `getSnapshotMatrix`. The same goes for `getSpatialBasis`.

tests/combine_samples.C
dylan-copeland: I see now that this string is used in `BasisGenerator`, so we can leave this as a string.

tests/combine_samples.C
jtlau: Added both verifications

tests/combine_samples.C
jtlau: It could be possible that the user doesn't want to subtract the mean or offset, so I've included that possibility here. Unless did you mean that I should include that verification inside this if statement?

tests/combine_samples.C
```diff
@@ -91,22 +91,26 @@ int main(int argc, char* argv[])
...
 	else {
             CAROM::Matrix *basis = (CAROM::Matrix*) reader.getSpatialBasis(0);
 	    dim    = basis->numRows();
 	    snaps += basis->numColumns();
+	    if (dimFirst == 0) dimFirst = dim;
 	}
     }
 	
     CAROM_VERIFY((snaps > 0) && (dim > 0));
+    CAROM_VERIFY(dim == dimFirst); // help to ensure files have the same dimensions
```
dylan-copeland, line 113: If this is inside the loop over `sample_names`, it will verify that all the files are of the same dimension.

tests/combine_samples.C
dylan-copeland: I meant the user should not specify both mean and offset.

jtlau: oh of course. Updated

tests/combine_samples.C
jtlau: I see, updated!

tests/combine_samples.C
jtlau: Done

jtlau:
The remaining two comments (currently) from @dylan-copeland  require a separate PR implementing a few smaller capabilities within libROM (a dimension read and editing of snapshots). I will start a new PR for those changes - they should not be complicated.

jtlau:
Finally got around the finishing this up.

The last PR I submitted created the functions `getDim` and `getNumSamples`, but only for snapshot matrix files. I generalized the functions for spatial and temporal bases just now (see latest commit).

Dylan, for your remaining suggestion, I briefly looked into adding the capability to modify existing snapshots, but there didn't seem to be a simple solution, so I agree that would be better for a future and separate PR. 

Can you look over the recent changes? If it looks good I believe this PR is finished.

tests/combine_samples.C
```diff
@@ -0,0 +1,190 @@
...
+    	}
+    	
+    	/*-- Load new samples with mean/offset subtracted --*/
+    	std::unique_ptr<CAROM::BasisGenerator> static_basis_generator2;
+    	static_basis_generator2.reset(new CAROM::BasisGenerator(
+    	    CAROM::Options(dim, snaps).setMaxBasisDimension(snaps), false,
+    	    generator_filename+"_sub"));
+    	
+    	for (int col = 0; col < num_cols; col++) {
+            CAROM::Vector *snap_cur = new CAROM::Vector(num_rows, true);
```
dylan-copeland, line 171: This allocates `snap_cur` for every column, without deleting it. It should be allocated before this loop over `col`. Also, it can be `CAROM::Vector snap_cur(num_rows, true);` so it doesn't have to be deleted.

tests/combine_samples.C
```diff
@@ -0,0 +1,190 @@
...
+        static_basis_generator->endSamples();
+    }
+    else {
+    	/*-- load data from hdf5 file to find the mean and subtract it --*/
+    	if (rank==0) std::cout << "Reading snapshots" << std::endl;
+    	CAROM::Matrix *snapshots = (CAROM::Matrix*) static_basis_generator->getSnapshotMatrix();
+        
+    	int num_rows = snapshots->numRows();
+    	int num_cols = snapshots->numColumns();
+        CAROM::Vector *offset = new CAROM::Vector(num_rows, true);
```
dylan-copeland, line 138: This could be `CAROM::Vector offset(num_rows, true);` rather than using `new` and `delete`.

dylan-copeland:
@jtlau Thanks for the recent changes. I agree that we can postpone subtracting an offset from a basis generator's snapshots to another PR. For now, it would be helpful to add a brief comment about how that should be implemented in the future, to eliminate the need for the second generator.

Also, the `combine_samples.C` should be in the `examples` directory, not `tests` (which is for regression tests). Maybe you can make a new subdirectory `examples/misc` for this file. This file is a general example, without concrete input/output that should be tested. It is not necessary, but if you wanted to make a test, this file could go in `examples`, and most of the code in `main` could be put into a function in the library which could be called from a test in the `tests` directory. Also not necessary, you could add comments on how to generate some snapshot data from one of the other examples in libROM.

Finally, the code needs to be stylized, but the script `stylize.sh` is broken. To fix it, this version can be copied and added to this PR:
https://github.com/LLNL/libROM/blob/save_dmd_csv/scripts/stylize.sh
If you are not familiar with astyle and would like me to do this and run astyle, just let me know and I will commit the formatting changes.

dylan-copeland:
@jtlau Thanks for the recent changes. I agree that we can postpone subtracting an offset from a basis generator's snapshots to another PR. For now, it would be helpful to add a brief comment about how that should be implemented in the future, to eliminate the need for the second generator.

Also, the `combine_samples.C` should be in the `examples` directory, not `tests` (which is for regression tests). Maybe you can make a new subdirectory `examples/misc` for this file. This file is a general example, without concrete input/output that should be tested. It is not necessary, but if you wanted to make a test, this file could go in `examples`, and most of the code in `main` could be put into a function in the library which could be called from a test in the `tests` directory. Also not necessary, you could add comments on how to generate some snapshot data from one of the other examples in libROM.

Finally, the code needs to be stylized, but the script `stylize.sh` is broken. To fix it, this version can be copied and added to this PR:
https://github.com/LLNL/libROM/blob/save_dmd_csv/scripts/stylize.sh
If you are not familiar with astyle and would like me to do this and run astyle, just let me know and I will commit the formatting changes.

jtlau:
Thanks for the comments @dylan-copeland. I moved the function to `examples`, and added its own CMakeLists section so it will be compiled even when not using MFEM. I was able to stylize the file, too. 

I created a test in `tests`, although it's not quite what you described, but I think it is a simple way to test the function and provide an example. What do you think?

dylan-copeland (approved):
Thanks @jtlau for addressing all my comments. This PR looks good!

siuwuncheung (approved):

chldkdtn (approved):

kevinhkhuynh (approved):

PR #99, opened 2022-02-24
Remove force update for macOS 

jtlau:
Small change to remove the forced update line for macOS. 

I recently did a clean install of librom and used `./scripts/compile.sh` (from the README), and my mac was forced to upgrade to Big Sur which took ~30 minutes I hadn't planned for. So, unless a software update is needed for some reason, I don't think this line is needed.


kevinhkhuynh:
Oops. Yes, this sounds like a good change to me.

PR #101, opened 2022-03-21
Fixing a bug in Matrix transpose

dylan-copeland:


siuwuncheung (approved):

chldkdtn (approved):

kevinhkhuynh:
Was transposing not working before or was it doing redundant work?

kevinhkhuynh (approved):

dylan-copeland:
> Was transposing not working before or was it doing redundant work?

It was looping over all entries, swapping A_ij with A_ji. This means the swapping was done twice, leaving the matrix unchanged. Only the upper (or lower) triangular part should be looped.

kevinhkhuynh:
> > Was transposing not working before or was it doing redundant work?
> 
> It was looping over all entries, swapping A_ij with A_ji. This means the swapping was done twice, leaving the matrix unchanged. Only the upper (or lower) triangular part should be looped.

There are a few different Matrix transpose unit tests in libROM. It is strange to me why they passed before the bug fix and continued to pass afterwards.

dylan-copeland:
@kevinhkhuynh In the `tests` directory, I see some tests for `transposeMult`, but not for `transpose`. Maybe a test should be added, if there is not one already.

PR #102, opened 2022-03-21
Get rows and columns from BasisReader

jtlau:
Add the capability to read only the dimension of the snapshot matrix, as suggested by @dylan-copeland in #98 .

Currently I've added that capability for just the snapshot matrix. I feel like this might be an opportunity to generalize the two added functions (`getDim` and `getNSamples`) for the basis and temporal matrices as well. Is there any need for that?

### DIFF
--- a/lib/linalg/BasisReader.h
+++ b/lib/linalg/BasisReader.h
@@ -293,7 +293,7 @@ public:
      * @return The number of samples in file.
      */
     int
-    getNSamples(
+    getNumSamples(
         double time);
 
     /**
--- a/lib/linalg/BasisReader.h
+++ b/lib/linalg/BasisReader.h
@@ -278,6 +278,26 @@ public:
 
     /**
      *
+     * @brief Returns the dimension of the system on this processor.
+     *
+     * @return The dimension of the system on this processor.
+     */
+    int
+    getDim(
+        double time);
+
+    /**
+     *
+     * @brief Returns the number of samples (columns) in file.
+     *
+     * @return The number of samples in file.
+     */
+    int
+    getNSamples(
+        double time);
+
+    /**
+     *
      * @brief Returns the snapshot matrix for the requested time.
      *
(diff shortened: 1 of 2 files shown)


lib/linalg/BasisReader.h
```diff
@@ -276,6 +276,26 @@ class BasisReader {
...
+        double time);
+
+    /**
+     *
+     * @brief Returns the number of samples (columns) in file.
+     *
+     * @return The number of samples in file.
+     */
+    int
+    getNSamples(
```
dylan-copeland, line 296: For consistency with other functions that use `Num`, could we rename this `getNumSamples`? It's hard to change names later, since it can break backwards compatibility.

tests/combine_samples.C
```diff
@@ -0,0 +1,198 @@
...
+    CAROM_VERIFY(!(subtract_mean && subtract_offset)); // both should not be selected 
+ 
+    /*-- Read dimension and count number of snapshots/bases --*/
+    if (rank==0) std::cout << "Opening files to read dimension and count number of snapshots/bases" << std::endl;
+    int dimFirst = 0;
+    for (const auto& sample_name: sample_names) {
+        CAROM::BasisReader reader(sample_name);
+		
+        if (kind == "snapshot") {
+	    CAROM::Matrix *snapshots = (CAROM::Matrix*) reader.getSnapshotMatrix(0);
```
dylan-copeland, line 101: This is the place to use the new `getDim` function, or did you want to add this file in the other PR?

jtlau: I was imagining finishing this PR first, but I realize this branch is built on top of the other changes. Let me drop those commits

jtlau:
This PR now only has the BasisReader changes (and does not include the `combine_samples.C` file).

dylan-copeland (approved):

chldkdtn (approved):

PR #107, opened 2022-05-03
Parametric DMD for CSV database

siuwuncheung:
`LIST_DIR/TRAIN_LIST.csv` and `LIST_DIR/TEST_LIST.csv`: 
- Both files are now assumed to have m rows and n+1 columns
- m is the number of {training, testing} parameter points
- n is the dimension of the parametric space
- first column is the name of the dataset `DATASET`, where the actual dataset is located in `DATA_DIR/DATASET`
- the second column through the last column denote n (could be zero) parameter values

`OUTPUT_DIR/indicator_val.csv`: 
- Optional. Present only for time windowing cases.
- It is now placed in the output directory, since it depends on the snapshot used and the model parameter `nwinsamp`. 
- It can be user-defined or auto-generated. 

`local_tw_csv.cpp`: 
- Renamed from `generic_csv.cpp`
- It is now limited to local (time-windowing) DMD. 
- Only allows one training parameter, i.e. only 1 row in `LIST_DIR/TRAIN_LIST.csv`.
- Allow multiple testing parameters predicted in one single run, i.e. multiple rows in `LIST_DIR/TEST_LIST.csv`.
- Ignore the parameters values, i.e. the second column through the last column of `LIST_DIR/TRAIN_LIST.csv` and `LIST_DIR/TEST_LIST.csv`.
- Prediction is made after training DMD models with `-train` (default) or use trained-and-saved models with `-no-train`. 
- As in the master branch, Indicator range partition and classification of snapshots is done simultaneously during the training of DMD. 

`parametric_tw_csv.cpp`:
- Newly added for parametric (time-windowing) DMD. 
- Offline and online phases are split.
- (Offline & online) Only allows multiple training parameters, i.e. more than 1 rows in `LIST_DIR/TRAIN_LIST.csv`.
- (Online) Allow multiple testing parameters predicted in one single run, i.e. multiple rows in `LIST_DIR/TEST_LIST.csv`.
- (Offline) Time windowing procedure: Indicator range partition is decided -> snapshots are classified into windows for each parameter -> training of DMD for each window and parameter. 
- (Online) Interpolating DMD models using the parameters values, i.e. the second column through the last column of `LIST_DIR/TRAIN_LIST.csv` and `LIST_DIR/TEST_LIST.csv`.
- (Offline & online) Prediction is made after training DMD models with `-predict` (default), on the training/testing datasets in the offline/online phase. 

Other changes: 
- Add an example script on heat conduction problem `heat_conduction_csv.sh` in `examples/dmd`
- Add results to the top of the DMD-CSV scripts `local_tw_csv.cpp` and `parametric_tw_csv.cpp` in `examples/dmd`


examples/dmd/generic_csv.cpp
```diff
@@ -155,190 +159,234 @@ int main(int argc, char *argv[])
...
+            }
+            dmd[window] = new CAROM::DMD(outputPath + "/window" + to_string(window)); 
         }
+    }
 
-        int overlap_count = 0;
-        for (int idx_snap = 0; idx_snap < num_train_snap[idx_dataset]; ++idx_snap)
+    StopWatch dmd_training_timer, dmd_preprocess_timer, dmd_prediction_timer;
+    double* sample = new double[dim];
+    CAROM::Vector* result = new CAROM::Vector(dim, true);
```
dylan-copeland, line 258: It looks like `result` is assigned to a pointer to a `Vector` returned by a function. Should it be allocated here?

examples/dmd/generic_csv.cpp
```diff
@@ -155,190 +159,234 @@ int main(int argc, char *argv[])
...
+        }
+        dmd_training_timer.Stop();
 
-        for (int window = 0; window < numWindows; ++window)
+        if (ddt > 0.0 && npar == 1) // TODO: Verify at all training par
         {
-            admd = dynamic_cast<CAROM::AdaptiveDMD*> (dmd[window]);
-            double t_init = dmd[window]->getTimeOffset();
+            CAROM::AdaptiveDMD* admd = nullptr; 
+            CAROM::Vector* interp_snap = new CAROM::Vector(dim, true);
```
dylan-copeland, line 342: Can this just be `CAROM::Vector interp_snap(dim, true);` so that it doesn't have to be deleted later?

examples/dmd/generic_csv.cpp
```diff
@@ -32,6 +31,7 @@
 #include <cmath>
 #include <iostream>
 #include <limits>
+#include <algorithm>
```
dylan-copeland, line 34: Is this needed? It builds without it.

examples/dmd/generic_csv.cpp
```diff
@@ -20,7 +20,6 @@
 // 5. DATA_DIR/DATASET/STATE/tval.csv     -- specifies the time instance of STATE
 // 6. DATA_DIR/dim.csv                    -- specifies the dimension of VAR_NAME
 // 7. DATA_DIR/index.csv                  -- (optional) each row specifies one DOF of VAR_NAME
-// 8. DATA_DIR/indicator_val.csv          -- (optional) each row specifies one indicator endpoint value
```
dylan-copeland, line 23: It would be helpful to have sample runs, as in `heat_conduction.cpp`.

examples/dmd/generic_csv.cpp
```diff
@@ -155,190 +159,234 @@ int main(int argc, char *argv[])
...
-                    new_window = (idx_snap >= (curr_window+1)*windowNumSamples);
+                    dmd[curr_window-1]->takeSample(sample, tval);
+                    overlap_count -= 1;
                 }
-                else 
-                {
-                    new_window = (tval >= indicator_val[curr_window+1]);
-                }
-                if (new_window)
+                if (curr_window+1 < numWindows && idx_snap+1 < num_train_snap[idx_dataset] &&
```
dylan-copeland, line 289: A comment about the condition here, involving the number 10, would be helpful.

siuwuncheung:
Thanks @dylan-copeland for reviewing. I will add more changes to this PR, so I remove RFR and add WIP back. I will review the code as suggested, before moving to RFR again.

examples/dmd/generic_csv.cpp
siuwuncheung: Thanks for the suggestion, Dylan. It is now allocated when assigned by the function.

examples/dmd/generic_csv.cpp
siuwuncheung: It is now deleted.

examples/dmd/generic_csv.cpp
siuwuncheung: Add a comment to `parametric_tw_dmd.csv`.

examples/dmd/generic_csv.cpp
siuwuncheung: Yes, they are now modified.

examples/dmd/generic_csv.cpp
siuwuncheung: For the purpose of adding sample runs, I have run the parametric heat conduction MFEM example and stored the outputs as CSV, so that it can be fed to the `generic_csv`. I planned to add the CSV files in `example/dmd`, but it is 700KB. I guess it is too large to be directly included into the repo?

chldkdtn: Perhaps, you can include generic_csv for a small number of time steps only.

siuwuncheung: Following the MFEM parametric heat conduction problem, 5 datasets, each with 4 time steps, are added.

siuwuncheung:
After some major changes, this PR is now ready for review.

chldkdtn (approved):

lib/utils/CSVDatabase.C
```diff
@@ -118,6 +118,24 @@ CSVDatabase::putComplexVector(
     d_fs.close();
 }
 
+void
+CSVDatabase::putStringVector(
+    const std::string& file_name,
+    const std::vector<std::string> data,
+    int nelements)
+{
+    CAROM_ASSERT(!file_name.empty());
```
dylan-copeland, line 127: `CAROM_ASSERT` -> `CAROM_VERIFY` will ensure this is checked even in optimized builds.

lib/utils/CSVDatabase.C
```diff
@@ -290,10 +308,15 @@ int
 CSVDatabase::getLineCount(
     const std::string& file_name)
 {
+    int count = 0;
```
dylan-copeland, line 311: `CAROM_VERIFY` (although this was not changed in this PR, it seems like a good idea to make it `VERIFY`).

examples/dmd/local_tw_csv.cpp
```diff
@@ -0,0 +1,640 @@
...
+            }
+            else // Verify DMD prediction results against dataset
+            {
+                while (curr_window+1 < numWindows && tval > indicator_val[curr_window+1])
+                {
+                    curr_window += 1;
+                }
+                if (myid == 0)
+                {
+                    cout << "Predicting DMD solution #" << idx_snap << " at t = " << tval << " using DMD model #" << curr_window << endl;
```
dylan-copeland, line 577: I don't think libROM has any policy about line length, but we generally try to avoid very long lines (e.g. over 100 characters). This file has several long lines that could be broken up.

examples/dmd/heat_conduction_csv.sh
```diff
@@ -0,0 +1,55 @@
...
+		  COMMAND="mpirun -oversubscribe"
+			MACHINE="Darwin";;
+    *)
+			echo "The regression tests can only run on Linux and MAC."
+			exit 1
+esac
+
+rm -rf parameters.txt hc_list hc_data run/hc_data
+
+$COMMAND -n 1 ./parametric_heat_conduction -r 0.40 -offline -rdim 16 -csv -out hc_data/hc_par1 > hc_par1.log
```
dylan-copeland, line 23: This script outputs many files to `build/examples/dmd`. It would be more convenient to write them to a subdirectory, e.g. `./out`, so that they can be removed easily.

examples/dmd/local_tw_csv.cpp
```diff
@@ -0,0 +1,640 @@
...
+//
+// User specify file locations and names by -list LIST_DIR -train-set TRAIN_LIST -test-set TEST_LIST -data DATA_DIR -var VAR_NAME -o OUT_DIR
+//
+// File structure:
+// 1. LIST_DIR/TRAIN_LIST.csv             -- each row specifies one training DATASET
+// 2. LIST_DIR/TEST_LIST.csv              -- each row specifies one testing DATASET
+// 3. LIST_DIR/DATASET.csv                -- each row specifies one STATE in DATASET
+// 4. DATA_DIR/dim.csv                    -- specifies the dimension of VAR_NAME
+// 5. DATA_DIR/DATASET/tval.csv           -- specifies the time instances
+// 6. DATA_DIR/DATASET/STATE/VAR_NAME.csv -- each row specifies one value of VAR_NAME of STATE
```
dylan-copeland, line 41: The solution field CSV files have one line for each unknown, written with 6 digits of accuracy. It seems that hdf5 output would be more accurate (more digits). The function `Vector::write` can be used.

examples/dmd/heat_conduction_csv.sh
```diff
@@ -0,0 +1,55 @@
...
+
+case "$(uname -s)" in
+    Linux*)
+		  COMMAND="srun -p pdebug"
+			MACHINE="Linux";;
+    Darwin*)
+		  COMMAND="mpirun -oversubscribe"
+			MACHINE="Darwin";;
+    *)
+			echo "The regression tests can only run on Linux and MAC."
```
kevinhkhuynh, line 17: This message can be changed.

kevinhkhuynh, line 17: This message can be changed.

examples/dmd/local_tw_csv.cpp
kevinhkhuynh: Another option is if you are using CSVDatabase to write out these files. You can instead increase the precision to 16 or some value equivalent to the HDF5Database. It is a one-liner.

https://stackoverflow.com/questions/5907031/printing-the-correct-number-of-decimal-points-with-cout

kevinhkhuynh: Another option is if you are using CSVDatabase to write out these files. You can instead increase the precision to 16 or some value equivalent to the HDF5Database. It is a one-liner.

https://stackoverflow.com/questions/5907031/printing-the-correct-number-of-decimal-points-with-cout

examples/dmd/local_tw_csv.cpp
kevinhkhuynh: It looks like astyle has an option to enforce line length. What do you think about using this option?

dylan-copeland: MFEM enforces 80-character lines, and I think it would be good to have a limit, if only to simplify reviewing (no more wondering if a line is too long, just make it automatic).

lib/utils/CSVDatabase.C
siuwuncheung: All the usage in CSVDatabase.C and HDFDatabase.C are changed to `CAROM_VERIFY`

examples/dmd/heat_conduction_csv.sh
siuwuncheung: The command line is changed to prevent unnecessary outputs.

examples/dmd/heat_conduction_csv.sh
```diff
@@ -14,7 +14,7 @@ case "$(uname -s)" in
 		  COMMAND="mpirun -oversubscribe"
 			MACHINE="Darwin";;
     *)
-			echo "The regression tests can only run on Linux and MAC."
+			echo "The CSV databse test can only run on Linux and MAC."
```
dylan-copeland: Typo

siuwuncheung: The line is now removed.

examples/dmd/local_tw_csv.cpp
siuwuncheung: I've added a 80-character limit to stylize, and it causes changes on few hundred lines. Is it okay to push it here when you are reviewing, or I should wait until you finish the review @dylan-copeland ?

dylan-copeland: For simplicity, I would prefer a separate PR just for the max line length. This PR can have long lines.

siuwuncheung: Sounds good to me.

examples/dmd/local_tw_csv.cpp
```diff
@@ -0,0 +1,640 @@
...
+            CAROM_VERIFY(snap_bound.size() == 2);
+            num_train_snap = snap_bound[1] - snap_bound[0] + 1;
+        }
+        else
+        {
+            num_train_snap = csv_db.getLineCount(string(list_dir) + "/" + par_dir + ".csv");
+        }
+
+        CAROM_VERIFY(windowOverlapSamples < windowNumSamples);
+        numWindows = (windowNumSamples < infty) ? ceil(num_train_snap / windowNumSamples) : 1;
```
dylan-copeland, line 253: I think this should be `ceil((double) num_train_snap / (double) windowNumSamples)`
Note that the output of 
cout << "ceil " << ceil(1.5) << ", " << ceil(3/2) << ", " << ceil((double) 3 / (double) 2) << endl;
is
ceil 2, 1, 2

examples/dmd/local_tw_csv.cpp
siuwuncheung: Set a higher precision of 10 for putDoubleArray and leave precision as a variable (default 6) in putDoubleVector and putComplexVector.

siuwuncheung:
@dylan-copeland has finished the review and informed me that the PR is ready to merge.

PR #108, opened 2022-05-04
Added toolchain

debog:
I have added the toolchain file required to compile on Quartz with GNU compilers, assuming they are currently loaded in the environment (i.e., the command "mpicc -v" or "mpicxx -v" should indicate a GNU compiler).

Also updated .gitignore to ignore docs/html

### DIFF
--- a/cmake/toolchains/gnu4-toss_3_x86_64_ib-librom-dev.cmake
+++ b/cmake/toolchains/gnu4-toss_3_x86_64_ib-librom-dev.cmake
@@ -1,0 +1,22 @@
+# Toolchain for Lawrence Livermore National Laboratory TOSS3 machines
+# (e.g., Quartz), assuming the following dependencies:
+#
+# - MVAPICH2/2.3 with Intel 19.0.4 compilers
+# - Intel MKL
+# - HDF5 1.8.18
+
+# Use MPI compiler wrappers because it simplifies detection of MPI
+set(CMAKE_C_COMPILER mpicc)
+set(CMAKE_CXX_COMPILER mpicxx)
+set(CMAKE_Fortran_COMPILER mpif90)
+set(BLA_VENDOR Intel10_64lp)
+set(HDF5_ROOT /usr/tce/packages/hdf5/hdf5-serial-1.8.17-gcc4.9.3)
+#string(APPEND CMAKE_CXX_FLAGS_INIT " -gxx-name=/usr/tce/packages/gcc/gcc-7.1.0/bin/g++")
+
+set(BLAS_ROOT /usr/tce/packages/mkl/mkl-2019.0/mkl)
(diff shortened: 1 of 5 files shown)


chldkdtn:
@kevinhkhuynh (https://github.com/LLNL/libROM/tree/master/scripts)/compile.sh file is dedicated to use intel compilers, right?

kevinhkhuynh:
> @kevinhkhuynh (https://github.com/LLNL/libROM/tree/master/scripts)/compile.sh file is dedicated to use intel compilers, right?

Any toolchain can be used with the -t option.

cmake/toolchains/gnu4-toss_3_x86_64_ib-librom-dev.cmake
```diff
@@ -0,0 +1,22 @@
+# Toolchain for Lawrence Livermore National Laboratory TOSS3 machines
+# (e.g., Quartz), assuming the following dependencies:
+#
+# - MVAPICH2/2.3 with Intel 19.0.4 compilers
+# - Intel MKL
+# - HDF5 1.8.18
+
+# Use MPI compiler wrappers because it simplifies detection of MPI
+set(CMAKE_C_COMPILER mpicc)
```
kevinhkhuynh, line 9: How about using the full path for the compilers so the toolchain file can work without doing "module load"?

cmake/toolchains/gnu4-toss_3_x86_64_ib-librom-dev.cmake
```diff
@@ -0,0 +1,22 @@
...
+set(CMAKE_CXX_COMPILER mpicxx)
+set(CMAKE_Fortran_COMPILER mpif90)
+set(BLA_VENDOR Intel10_64lp)
+set(HDF5_ROOT /usr/tce/packages/hdf5/hdf5-serial-1.8.17-gcc4.9.3)
```
debog, line 13: @kevinhkhuynh @chldkdtn I am unhappy with the HDF5_ROOT line in this toolchain. I set the compilers to what is currently loaded, but the HDF5_ROOT is hard-coded to the one compiled by gcc4.9.3, which is okay for me since that's the compiler that's loaded in my environment.

Is there a way to automate this to the HDF5 library currently loaded?

For example, the bash equivalent would be something like:
```
H5DIFF_SUFFIX=/bin/h5diff
H5DIFF_PATH=$(which h5diff)
export HDF5_ROOT=${H5DIFF_PATH%$H5DIFF_SUFFIX}

```

debog: Resolved!

kevinhkhuynh (approved):

kevinhkhuynh:
Thanks @debog looks good. I just made a quick change to use the default toolchain in the compile script.

siuwuncheung (approved):

PR #116, opened 2022-06-02
Lwork overflow bug fix

kevinhkhuynh:
bestwork is such a large number that lwork is overflowing, leading to a negative value, which messes up the allocate for the work array, leading to a seg fault.

Without changing to another SVD algorithm, the best we can do is allow lwork to be at max MAX_INT.

### DIFF
--- a/lib/linalg/scalapack_f_wrapper.f90
+++ b/lib/linalg/scalapack_f_wrapper.f90
@@ -23,8 +23,6 @@ module scalapack_wrapper
 
     integer, parameter :: REAL_KIND = C_DOUBLE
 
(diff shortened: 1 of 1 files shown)


kevinhkhuynh:
Creating a custom long int parameter and setting lwork to that failed, since pdgesvd and the other lapack algorithms require lwork to be an INTEGER(4), and setting it to something like INTEGER(8) leads to a compilation failure.

kevinhkhuynh:
This bug fix might lead to incorrect results when the lwork is so large that setting it to MAX_INT, leads to not enough workspace, and bad results. We could instead output a Fortran error, telling users the matrix is too large to work with, instead of seg faulting.

PR #121, opened 2022-06-10
updated mixed nonlinear diffusion

kevinhkhuynh:
Update mixed nonlinear diffusion

### DIFF
--- a/examples/prom/mixed_nonlinear_diffusion.cpp
+++ b/examples/prom/mixed_nonlinear_diffusion.cpp
@@ -63,6 +63,7 @@
 #include "linalg/BasisGenerator.h"
 #include "linalg/BasisReader.h"
 #include "hyperreduction/DEIM.h"
+#include "hyperreduction/GNAT.h"
 #include "hyperreduction/S_OPT.h"
 #include "mfem/SampleMesh.hpp"
 
@@ -70,8 +71,6 @@
 typedef enum {ANALYTIC, INIT_STEP} PROBLEM;
 
 typedef enum {RSPACE, WSPACE} FESPACE;
-
-//#define USE_GNAT
 
 using namespace mfem;
 using namespace std;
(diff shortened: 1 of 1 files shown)


examples/prom/mixed_nonlinear_diffusion.cpp
```diff
@@ -533,6 +534,8 @@ int main(int argc, char *argv[])
                    "Enable or disable the merge phase.");
     args.AddOption(&use_sopt, "-sopt", "--sopt", "-no-sopt", "--no-sopt",
```
dylan-copeland, line 535: This is not a change in this PR, but I noticed that `--sopt` is in the sample runs, which could be `-sopt`.

examples/prom/mixed_nonlinear_diffusion.cpp
```diff
@@ -1602,7 +1615,7 @@ RomOperator::RomOperator(NonlinearDiffusionOperator *fom_,
                          const CAROM::Matrix *Bsinv,
                          const double newton_rel_tol, const double newton_abs_tol, const int newton_iter,
                          const CAROM::Matrix* S_, const CAROM::Matrix *Ssinv_,
-                         const int myid, const bool hyperreduce_source_)
+                         const int myid, const bool hyperreduce_source_, const bool oversampling_)
```
dylan-copeland, line 1618: It is awkward having to use this variable `oversampling` just to indicate that `Bsinv` should be transposed when multiplying. The alternative of allocating the transpose is inefficient. Maybe a better design would be to add a `transpose` data member in the `Matrix` class, indicating that `transposeMult` should be called for `mult` and `mult` should be called for `transposeMult`. I'm ok with leaving this PR as it is, but if we see a frequent need for this transpose logic, we may want to consider putting the logic in the `Matrix` class.

dylan-copeland (approved):

examples/prom/mixed_nonlinear_diffusion.cpp
kevinhkhuynh: Hi Dylan. Sounds good to me. Let's leave it like this for now and return back to the issue if it becomes a recurrent issue.

PR #123, opened 2022-06-15
DMD with shifted data

siuwuncheung:
In junction with [Laghos PR#171](https://github.com/CEED/Laghos/pull/171).

This PR adds the capability of shifting the snapshot data in DMD using offsets and predicting derivatives. Corresponding prediction formula is implemented. 

Although the motivation of adding this capability is to work on the Laghos examples with zero initial conditions in velocity (triple point expansion and Sedov blast explosion), a few test runs show that these new capabilities do not provide sufficiently accurate predictions yet. In the future, more investigations are needed on Laghos and these capabilities shall be applied to other MFEM examples, but they are not urgent.


dylan-copeland:
Are there any examples in libROM, or just Laghos?

lib/algo/DMD.h
```diff
@@ -156,8 +166,13 @@ class DMD
      * @brief Constructor.
      *
      * @param[in] dim        The full-order state dimension.
+     * @param[in] init_os_s  Use initial state as offset.
+     * @param[in] init_os_d  Use initial derivative as offset.
+     * @param[in] mean_os_s  Use state mean as offset.
+     * @param[in] mean_os_d  Use derivative mean as offset.
```
dylan-copeland, line 172: It seems a state offset and a derivative offset can be used together. This should be explained in a comment.

lib/algo/DMD.h
```diff
@@ -258,6 +283,36 @@ class DMD
      */
     std::vector<Vector*> d_sampled_times;
 
+    /**
+     * @brief Whether to use init state as offset.
+     */
+    bool d_init_os_s = false;
+
+    /**
+     * @brief Whether to use init derivative as offset in NonuniformDMD.
```
dylan-copeland, line 292: Should `NonuniformDMD` be mentioned in the base class?

lib/algo/DMD.cpp
```diff
@@ -343,11 +395,98 @@ DMD::constructDMD(const Matrix* f_snapshots,
...
+        // Find which columns of d_basis are linearly independent from W0
+        for (int j = 0; j < d_basis->numColumns(); j++)
+        {
+            // l = W0' * u
+            Vector W_col(f_snapshots->numRows(), f_snapshots->distributed());
+            for (int i = 0; i < f_snapshots->numRows(); i++)
+            {
+                W_col.item(i) = d_basis->item(i, j);
+            }
+            Vector* l = W0->transposeMult(W_col);
```
dylan-copeland, line 416: This allocates `l` for each `j`. It would be more efficient to construct a work vector before the loop. Also for `W0l`.

lib/algo/DMD.cpp
```diff
@@ -490,6 +630,15 @@ DMD::predict(double t)
     Vector* d_predicted_state_real = d_predicted_state_real_1->minus(
                                          d_predicted_state_real_2);
 
+    if (d_state_offset)
+    {
+        *d_predicted_state_real += *d_state_offset;
+    }
+    if (d_derivative_offset)
+    {
+        *d_predicted_state_real += *(d_derivative_offset->mult(t));
```
dylan-copeland, line 639: It looks like this `mult` allocates a vector. Can this be done without a vector allocation?

lib/algo/DMD.h
siuwuncheung: This is now replaced by a general offset option.

lib/algo/DMD.h
siuwuncheung: Both AdaptiveDMD and Nonuniform DMD are derived classes of DMD, and there are some options/variables that only apply to AdaptiveDMD (and vanilla DMD) or Nonuniform DMD. I think adding some comments to distinguish them will make good references to further development/usage.

lib/algo/DMD.cpp
siuwuncheung: Added.

lib/algo/DMD.h
dylan-copeland: If there is a way to avoid mentioning derived classes in the base class, I think that would be preferable. A variable should be in the base class only if it is potentially used in general for multiple derived classes. If it is only intended for one derived class, as this comment seems to suggest, then it should be a member of the derived class, not the base class. Is there a better way to design this, by putting the variable in the derived class? Or is it a general variable, so that there should be no mention of `NonuniformDMD`, i.e. just comment `Whether to use init derivative as offset.`

siuwuncheung: Now `d_derivative_offset` is only in `NonuniformDMD` class, and a new function `addOffset` is added.

lib/algo/DMD.cpp
siuwuncheung: It is replaced by `plusAx` now.

siuwuncheung:
> Are there any examples in libROM, or just Laghos?

It is just used in Laghos now. In a future PR I will add the usage to other MFEM examples.

lib/algo/NonuniformDMD.h
```diff
@@ -2,8 +2,6 @@
  *
  * Copyright (c) 2013-2022, Lawrence Livermore National Security, LLC
  * and other libROM project developers. See the top-level COPYRIGHT
- * file for details.
```
dylan-copeland, line 5: Were these 2 lines intended to be deleted?

lib/algo/DMD.cpp
```diff
@@ -56,12 +58,14 @@ DMD::DMD(int dim)
...
     d_dim = dim;
+    d_state_offset = state_offset;
     d_trained = false;
     d_init_projected = false;
 }
 
-DMD::DMD(int dim, double dt)
+DMD::DMD(int dim, double dt, Vector* state_offset)
 {
+    // Vanilla DMD
```
dylan-copeland, line 68: Although we know what "vanilla" means here, users may not. Can you make a more descriptive comment that this is the basic DMD with uniform steps?

lib/algo/DMD.cpp
```diff
@@ -343,11 +357,112 @@ DMD::constructDMD(const Matrix* f_snapshots,
...
+            }
+
+            // Use k to see if the vector addressed by u is linearly dependent
+            // on the left singular vectors.
+            if (k >= linearity_tol)
+            {
+                lin_independent_cols_W.push_back(j);
+            }
+            delete d_basis_init;
+            delete l;
```
dylan-copeland, line 418: Instead of allocating and deleting vectors in each iteration, can work vectors be allocated before the loop? If the vectors are allocated in the multiplications, is there another version of the multiplication that uses a vector already allocated?

lib/algo/DMD.cpp
siuwuncheung: The comments are modified and now put in the header file.

lib/algo/NonuniformDMD.h
siuwuncheung: Thanks for pointing that out. It was an accident.

lib/algo/DMD.cpp
siuwuncheung: They are modified now.

chldkdtn (approved):

dylan-copeland (approved):

siuwuncheung:
LGTM

PR #136, opened 2022-08-09
Example: Nonlinear Elastic global pROM

axla-io:
# Example: Nonlinear Elastic Global pROM

## Background and description
To extend the `libROM `library, I implemented a global parametric ROM for a nonlinear elastic cantilever problem.

The program is based on [ex10p](https://github.com/mfem/mfem/blob/master/examples/ex10p.cpp) in MFEM and the `mixed_nonlinear_diffusion` example in `libROM`.

## What's added

The file:
* `examples/prom/nonlinear_elasticity_global_rom.cpp` - which is the main program for the pROM example

This example adds:
* Hyper reduction for nonlinear elastic problems.
* The option to chose approximating the basis for the velocity field using the basis for the displacement field.
* The possibility to chose velocity or deformation initial conditions.
* Parametric ROM mode by scaling initial conditions.
* BasisGeneratorFinalSummary generates .txt files, which are easier to read on Windows.

## How has it been tested?
* Hyper reduction implementation verified.
* Parametric ROM interpolation capabilities verified.
* X basis approach implementation has shown to give similar performance as using both bases.


CMakeLists.txt
```diff
@@ -132,12 +132,12 @@ if (USE_MFEM)
     dg_advection_global_rom
     dg_advection_local_rom_matrix_interp
     mixed_nonlinear_diffusion
+    nonlinear_elasticity_global_rom
     linear_elasticity_global_rom
     dg_advection
     nonlinear_elasticity
     heat_conduction
     parametric_heat_conduction
-    de_parametric_heat_conduction_greedy
```
dylan-copeland, line 140: It looks like this was removed by mistake.

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+#include "hyperreduction/S_OPT.h"
+#include "mfem/SampleMesh.hpp"
+
+#include <memory>
+#include <cmath>
+#include <limits>
+#include <iostream>
+#include <fstream>
+
+typedef enum {FSPACE } FESPACE;
```
dylan-copeland, line 100: This is not used.

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+    bool def_ic = false;
+
+    // ROM parameters
+    bool offline = false;
+    bool merge = false;
+    bool online = false;
+    bool use_sopt = false;
+    bool hyperreduce = true;
+    bool x_base_only = false;
+    int num_samples_req = -1; // 1170 for full sampling
```
dylan-copeland, line 395: This problem-specific comment should be removed.

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+    BlockVector vx0 = BlockVector(vx);
+    BlockVector vx_diff = BlockVector(vx);
+
+    // Reduced order solution
+    Vector* wMFEM = 0;
+    CAROM::Vector* w = 0;
+    CAROM::Vector* w_v = 0;
+    CAROM::Vector* w_x = 0;
+
+
```
dylan-copeland, line 664: Extra line

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+            // Count number of true elements in new matrix
+            int num_ess_sp = 0;
+
+            for (size_t i = 0; i < sp_size; i++)
+            {
+                if (Ess_mat_sp(i,0) == 1)
+                {
+                    num_ess_sp += 1;
+                }
+
```
dylan-copeland, line 990: Extra line

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+            // Add indices to list
+            int ctr = 0;
+            for (size_t i = 0; i < sp_size; i++)
+            {
+                if (Ess_mat_sp(i,0) == 1)
+                {
+                    ess_tdof_list_sp[ctr] = i;
+                    ctr += 1;
+                }
+
```
dylan-copeland, line 1005: Extra line

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+            }
+
+            // Define operator in sample space
+            soper = new HyperelasticOperator(*sp_XV_space, ess_tdof_list_sp, visc, mu, K);
+
```
dylan-copeland, line 1010: Extra line

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+                                    num_samples_req != -1, hyperreduce, x_base_only);
+        }
+        else
+        {
+            romop = new RomOperator(&oper, soper, rvdim, rxdim, hdim, smm, vx0.GetBlock(0),
+                                    vx0.GetBlock(1), vx0.GetBlock(0), BV_librom, BX_librom, H_librom, Hsinv, myid,
+                                    num_samples_req != -1, hyperreduce, x_base_only);
+        }
+
+        // Display lifted initial energies
```
dylan-copeland, line 1025: Would `Print` be more accurate than `Display`? `Display` sounds like a visualization is being done.

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+        {
+            if (myid == 0)
+            {
+                solveTimer.Start();
+                ode_solver->Step(*wMFEM, t, dt_real);
+                solveTimer.Stop();
+            }
+
+            MPI_Bcast(&t, 1, MPI_DOUBLE, 0, MPI_COMM_WORLD);
+
```
dylan-copeland, line 1083: Extra line (and some more below)

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+            solveTimer.Stop();
+
+        }
+
+        last_step = (t >= t_final - 1e-8 * dt);
+
+        if (offline)
+        {
+
+            if (  basis_generator_x->isNextSample(t) || x_base_only == false && basis_generator_v->isNextSample(t))
```
dylan-copeland, line 1100: Extra spaces. Astyle may fix this automatically.

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+
+        // Take samples
+        if (x_base_only == false)
+        {
+        basis_generator_v->takeSample(vx_diff.GetBlock(0), t, dt);
+        basis_generator_v->writeSnapshot();
+        delete basis_generator_v;
+        }
+        
+        //Ändra bug!
```
dylan-copeland, line 1214: Remove?

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+    // 16. Free the used memory.
+    delete ode_solver;
+    delete pmesh;
+
+    totalTimer.Stop();
+    if (myid == 0) cout << "Elapsed time for entire simulation " <<
+                            totalTimer.RealTime() << endl;
+
+    MPI_Finalize();
+    return 1;
```
dylan-copeland, line 1315: It may not really matter, but we usually have `return 0` for a successful run.

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1747 @@
...
+    fom->Smat.Mult(v0_fom, *S_hat_v0_temp);
+    V_v.transposeMult(*S_hat_v0_temp_librom, S_hat_v0);
+
+    // Create M_hat
+    Compute_CtAB(fom->Mmat, V_v, V_v, M_hat);
+
+    // Invert M_hat and store
+    M_hat->inverse(*M_hat_inv);
+
+    //if (myid == 0)
```
dylan-copeland, line 1545: Should `myid == 0` be used or removed? Has this example been tested in parallel?

chldkdtn:
@dylan-copeland Do you know what enables the mesh deformation visualization in VisIt? Currently, it seems that the mesh is not deformed in VisIt for this particular code.

dylan-copeland:
> @dylan-copeland Do you know what enables the mesh deformation visualization in VisIt? Currently, it seems that the mesh is not deformed in VisIt for this particular code.

The mesh has to have its nodes updated in the simulation, as done in Laghos, for example LagrangianHydroOperator::UpdateMesh in laghos_solver.cpp. Otherwise, the original mesh is output by the simulation and read by VisIt.

chldkdtn:
> 

Yes, but the thing is, it looks like this code is already updating nodes by the following lines:

        GridFunction* nodes = &x_gf;
        int owns_nodes = 0; 
        pmesh->SwapNodes(nodes, owns_nodes);

chldkdtn:
> 

@copeland, no worries, it works now. the part I wrote above needs to be inside of the time loop. Thank you for your help. :)

examples/prom/nonlinear_elasticity_global_rom.cpp
chldkdtn: @axla-io I have just tested it in parallel and found that it crashes.

CMakeLists.txt
axla-io: Yes that is most likely!

examples/prom/nonlinear_elasticity_global_rom.cpp
axla-io: You are right! I used it in the debugging phase and it stuck

examples/prom/nonlinear_elasticity_global_rom.cpp
axla-io: That's true. I'll run Astyle again

examples/prom/nonlinear_elasticity_global_rom.cpp
axla-io: Thank you for spotting all these errors Dylan!

examples/prom/nonlinear_elasticity_global_rom.cpp
axla-io: @chldkdtn @dylan-copeland That is disturbing. I think that I tried it in parallel with my two processors, but I'm not sure. I will try it on my side and do some debugging.

I'll also check whether the `myid == 0` should be there.

examples/prom/nonlinear_elasticity_global_rom.cpp
axla-io: I agree.

examples/prom/nonlinear_elasticity_global_rom.cpp
axla-io: @chldkdtn Just to be clear, does the `offline` and `merge` modes work for you? They do for me, but I get a crash on the `online` mode, which I'll debug.

axla-io: @chldkdtn @dylan-copeland 
I've fixed the MPI issue for the online mode in the latest commit. It seems, however, that I get into a race condition when I use MPI for the merge mode with more than two snapshots (for parametric ROM). This is surprising to me as I followed the `mixed_nonlinear_diffusion` code closely. Could you take a look and see if you also get a race condition with the latest commit if you run:

```
mpirun -np 2 nonlinear_elasticity_global_rom --offline -dt 0.01 -tf 5.0 -s 14 -vs 100 -sc 0.90 -xbo -def-ic -id 0

mpirun -np 2 nonlinear_elasticity_global_rom --offline -dt 0.01 -tf 5.0 -s 14 -vs 100 -sc 1.10 -xbo -def-ic -id 1

mpirun -np 2 nonlinear_elasticity_global_rom --merge -ns 2 -dt 0.01 -tf 5.0 -xbo
```

Oddly enough, running
```
mpirun -np 2 nonlinear_elasticity_global_rom --merge -ns 1 -dt 0.01 -tf 5.0 -xbo
```

Works for me, so maybe it's just that my computer is slow? It could also be that the `MergeBasis` function from `mixed_nonlinear_diffusion` is not made to handle parametric ROMs, but that's just speculation.

axla-io: Also, it turns out that there is some difference for the hyper reduced results between parallel and serial mode. I think that we need to understand whether this is connected to the merge issue as well before looking into it more.

dylan-copeland: The `MergeBasis` function in `mixed_nonlinear_diffusion` works for parametric ROM in serial. There is a serial sample run that works, to verify this. There may be multiple reasons why that test does not work in parallel, and I'm not if it involves the merge phase. I ran the above sample run for `nonlinear_elasticity_global_rom` and also found that it hangs. At first glance, I don't see what the issue is. If it is difficult to debug, I would say that getting this PR to work in serial is good enough, since other examples in the `prom` directory do not work in parallel yet.

axla-io: I understand. For the parametric ROM, I should have precised that I meant that the `mixed_nonlinear_diffusion` example maybe couldn't handle parametric ROMs in parallel.

But then, for now this is good enough now and more effort can be put into the parallelization at a later stage? Maybe making the snapshot writing, and base merging serial could solve our issues?

dylan-copeland: In general, snapshot writing and basis merging should be parallel. Laghos does these things in parallel. It is important to keep the data distributed, since in theory large problems could have so much data that memory limits are exceeded in serial, not to mention the performance bottleneck. So it should work in parallel, but it has to be debugged. For now, I am fine with leaving this for future work.

axla-io: Thanks Dylan, that makes a lot of sense. I guess there is also a bottleneck in the implementation here. In `mixed_linear_diffusion`, the ROM operator only works in serial in the online mode, same goes for this program. So in order to have parallelization we need to change the implementation of the ROM online mode as well.

chldkdtn: @axla-io is it possible to make the parallel version work in this PR? That would be wonderful.

dylan-copeland: @axla-io The fact that `mixed_nonlinear_diffusion` only works in serial is due to bugs, and the question of whether the online stage should be serial or parallel is separate but important. In that example, you can see that `ode_solver.Step` is called in the online case only for rank 0. The idea is that the online stage must be run in parallel in order to load the parallel basis files (created by the merge stage), but the ROM problem is too small to parallelize. Ideally, we would have another stage as in Laghos for preprocessing in parallel everything that is needed for the serial online stage and writing it to file. For example, the preprocessing stage would load the basis files in parallel and compute the reduced matrices and write them to file, to be read by the serial online stage. That level of optimization is not really necessary in these examples.

Also, to clarify, I think it would be acceptable to have this example work only in serial for this PR, but of course it is worth trying to fix the parallel case if it doesn't take too much work.

chldkdtn: @axla-io Let's not worry about parallel implementation in this PR as @dylan-copeland said.

axla-io: Thanks for the explanation @dylan-copeland !

axla-io: @chldkdtn Okay, that sounds good. I'll turn my focus to the report writing now.

chldkdtn (approved):

examples/prom/nonlinear_elasticity_global_rom.cpp
```diff
@@ -0,0 +1,1749 @@
...
+//      ./nonlinear_elasticity_global_rom --offline -dt 0.01 -tf 5.0 -s 14 -vs 100 -sc 1.00 -id 2
+//
+// Online phase with full sampling:
+//      ./nonlinear_elasticity_global_rom --online -dt 0.01 -tf 5.0 -s 14 -vs 100 -hyp -rvdim 40 -rxdim 10 -hdim 71 -nsr 1170 -sc 1.00
+// Output message:
+//      Elapsed time for time integration loop 14.0381
+//      Relative error of ROM position (x) at t_final: 5 is 2.97882e-06
+//      Relative error of ROM velocity (v) at t_final: 5 is 0.000679751
+//
+// Online phase with strong hyper reduction:
```
dylan-copeland, line 50: This is indeed strong hyperreduction. With `nsr 100` the speedup is still good, without about 3% error. Maybe a brief comment about that would be helpful. Otherwise, the reader has to experiment to find this.

dylan-copeland:
@axla-io I will be ready to approve this PR once `scripts/stylize.sh` is run. Please note that the master branch has been updated, so it would be best to pull from master first.

examples/prom/nonlinear_elasticity_global_rom.cpp
chldkdtn: I have included -nsr 100 case below. I have also did astyle. @dylan-copeland if you can review one more time, that would be wonderful.

axla-io: Hi @chldkdtn @dylan-copeland Sorry I've been unresponsive for a couple of days. A lot of school work unfortunately. Thank you, @chldkdtn for adding the example and doing astyle. Is there anything more I can help out with?

dylan-copeland (approved):

PR #143, opened 2022-09-24
Pointwise Snapshots in MFEM

dylan-copeland:
This PR adds the capability to output snapshots of pointwise solution values on a grid of physical points, for MFEM fields. The example `mixed_nonlinear_diffusion` demonstrates the capability.


chldkdtn:
@pranav-vempati is compilation option with gslib added?

pranav-vempati:
The option is added but there are still some compilation issues with the code that need to be resolved.

chldkdtn:
@pranav-vempati Is it still not working?

pranav-vempati:
@chldkdtn I believe that the underlying error has been fixed with my latest commit. I'll push a commit soon that cleans up some of the code.

pranav-vempati (approved):

chldkdtn:
I did "compile.sh -mg" to compile libROM and tried to run the following:

./mixed_nonlinear_diffusion -pwsnap -pwx 101 -pwy 101

in build/examples/prom/ directory, but I get the following error:


To use pointwise snapshots, compile with -mg option


MFEM abort: Pointwise snapshots aren't available, since the compilation is done without the -mg option
 ... in function: int main(int, char **)
 ... in file: /usr/WS1/choi15/libROM/examples/prom/mixed_nonlinear_diffusion.cpp:694

Who should look into this error? @pranav-vempati or @dylan-copeland ?

dylan-copeland:
@chldkdtn It may be necessary for you to remove everything librom/dependencies, except parmetis-4.0.3.tar.gz, in order for MFEM to be built with gslib. If an MFEM build is already present, it will not be replaced.

chldkdtn:
> @chldkdtn It may be necessary for you to remove everything in librom/dependencies, except parmetis-4.0.3.tar.gz, in order for MFEM to be built with gslib. If an MFEM build is already present, it will not be replaced.

You are right, @dylan-copeland . After deleting everything except parmetis-4.0.3.tar.gz, I was able to run pointwise snapshots examples. By the way, we have to fix this. @pranav-vempati , is it possible to detect if MFEM build is already present when -mg option is used, then erase everything in librom/dependencies except parmetis-4.0.3.tar.gz, and then do the compiling?

chldkdtn:
@dylan-copeland it is a bit odd that point wise snapshots example is in prom folder although it is mainly for DMD simulations. Can you introduce point wise snapshots example in dmd folder?

dylan-copeland:
@pranav-vempati For the cleanup, isn't it sufficient to remove only `dependencies/mfem*`, i.e. all the mfem builds? I don't think the other libraries (hypre, parmetis, etc) need to be rebuilt for the `-mg` case. Also, it would be more efficient  to remove the mfem builds only if necessary. If a previous build was with `-mg`, and the current build is also with `-mg`, it is not necessary to clean the dependencies. Can a configuration file store the previous build options, or can the script check the configuration file in mfem to see whether the gslib option is on? I think it is acceptable to keep the old build also if it was done with `-mg` and the current build is `-m`.

dylan-copeland:
@chldkdtn I added the snapshot output option in the dmd example `parametric_heat_conduction`.

pranav-vempati:
@dylan-copeland I'm checking for the presence of the gslib folder to determine if the mfem dependencies need to be removed. If the -mg option is used, and there is no gslib folder, I assume that the previous compilation was with -m, and delete the mfem libraries prior to compilation. Is this ok?

dylan-copeland:
> @dylan-copeland I'm checking for the presence of the gslib folder to determine if the mfem dependencies need to be removed. If the -mg option is used, and there is no gslib folder, I assume that the previous compilation was with -m, and delete the mfem libraries prior to compilation. Is this ok?

Thanks @pranav-vempati, this sounds good to me. It is possible that a user builds with `-mg`, removes mfem manually, and then builds with `-m` followed by a `-mg` build. In that case, the user should know what they are doing and should manually delete mfem. Also, I don't see a reason to build without gslib after building with gslib, so we can neglect such cases.

chldkdtn (approved):

siuwuncheung (approved):

PR #150, opened 2022-10-22
BasisGenerator finish output on function call

dylan-copeland:
Before this PR, `BasisGenerator` does not finish writing out basis or snapshot files, when `endSamples` or `writeSnapshot` is called. The files are fully written only when `BasisGenerator` is deleted or goes out of scope, as the writer has to be deleted. This PR deletes the writer when `endSamples` or `writeSnapshot` is called, so that the files are finished. This prevents confusion and bugs, and the only disadvantage is that `endSamples` or `writeSnapshot` cannot be called twice.

### DIFF
--- a/lib/linalg/BasisGenerator.h
+++ b/lib/linalg/BasisGenerator.h
@@ -128,6 +128,8 @@ public:
     {
         if (d_basis_writer) {
             d_basis_writer->writeBasis(kind);
+            delete d_basis_writer;
+            d_basis_writer = nullptr;
         }
     }
 
@@ -139,6 +141,8 @@ public:
     {
         if (d_basis_writer) {
             d_basis_writer->writeBasis("snapshot");
+            delete d_basis_writer;
+            d_basis_writer = nullptr;
         }
     }
 
--- a/lib/linalg/BasisGenerator.h
+++ b/lib/linalg/BasisGenerator.h
@@ -131,6 +131,11 @@ public:
             delete d_basis_writer;
             d_basis_writer = nullptr;
         }
+        else
+        {
+            std::cout << "WARNING: file has already been written by endSamples"
+                      << std::endl;
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jtlau:
Is it still possible to write both the snapshot and basis files?

I see the need for this. In `tests/smoke_static.C` I created an inner scope just so the files would be closed in the destructor. And I probably knew to do that only after some confusion. 

But I also see in that example that `writeSnapshot()` and `endSamples()` are both called. I don't call both in practice so it might just be this testing example.

dylan-copeland:
@jtlau Thanks for watching this PR. I was planning to ask whether this PR might break something, as it makes it impossible to call `endSamples` twice, or to call both `endSamples` and `writeSnapshot`. This is why it is a WIP. It is helpful to know that you do not see a practical use for calling both, except in a unit test. I will test Laghos and our other examples with this PR to see if anything is broken, and of course anyone who sees an issue with this PR should comment. I can also add an error message if `endSamples` or `writeSnapshot` is called after the writer is deleted, since an error message would be more helpful than the current version that simply fails.

chldkdtn:
@dylan-copeland what is the status of this PR?

dylan-copeland:
> @dylan-copeland what is the status of this PR?

It turned out that the basis generators are more complex than I first anticipated, since they support time intervals, where new intervals are made when the maximum number of samples is taken in an interval. Because of this, I did not have time or motivation yet to come up with a plan. Maybe we should consider removing the interval feature, if no one ever uses it, and it only causes unnecessary complications.

chldkdtn:
> > @dylan-copeland what is the status of this PR?
> 
> It turned out that the basis generators are more complex than I first anticipated, since they support time intervals, where new intervals are made when the maximum number of samples is taken in an interval. Because of this, I did not have time or motivation yet to come up with a plan. Maybe we should consider removing the interval feature, if no one ever uses it, and it only causes unnecessary complications.

I cast my vote to remove the interval feature. I do not think we use it anywhere.

dylan-copeland:
> I cast my vote to remove the interval feature. I do not think we use it anywhere.

I agree. We should remove that feature first in a separate PR, and then revisit this PR, which will be greatly simplified.

chldkdtn:
@dylan-copeland The plan is to remove interval concept in basis generator and then come back to this PR again.

dreamer2368:
This work is addressed by #261.

PR #153, opened 2022-11-02
HDF DMD files

dylan-copeland:
This PR implements the option to generate binary HDF files for the solutions of the `dmd/parametric_heat_conduction.cpp` example. Previously, CSV files were the only option. CSV is still used for some small amounts of data describing the simulations, but HDF can be used for solution data. The advantage of the HDF option is that the data is stored in full precision more efficiently than CSV, and there is only 1 HDF file per simulation as opposed to one CSV file per timestep. For example, `heat_conduction_hdf.sh` runs several times faster than `heat_conduction_csv.sh` for generating training data. The `Database` classes are enhanced to support the new features.


chldkdtn (approved):

chldkdtn:
@siuwuncheung @pranav-vempati let's review and approve this PR.

siuwuncheung (approved):

pranav-vempati (approved):

PR #157, opened 2022-11-09
Updating overlap method in Parametric DW-DMD

siuwuncheung:
By specifying `rdim`, we can now make sure each DMD model collects at least `rdim` snapshots.

### DIFF
--- a/examples/dmd/parametric_dw_csv.cpp
+++ b/examples/dmd/parametric_dw_csv.cpp
@@ -164,11 +164,6 @@ int main(int argc, char *argv[])
     CAROM_VERIFY(!(offline && online) && (offline || online));
     CAROM_VERIFY(!(dtc > 0.0 && ddt > 0.0));
     CAROM_VERIFY(numWindows == 1 || windowNumSamples == infty);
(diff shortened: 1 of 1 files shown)


pranav-vempati:
Hi Tony, is this PR ready for review?

siuwuncheung:
> Hi Tony, is this PR ready for review?

Yes, it is RFR now. Again, the scripts `local_dw_csv` and `parametric_dw_csv` are designed only for DMD on the pore collapse data now, so there is no testing examples.

dylan-copeland (approved):
I am approving since there are no example runs and nothing to verify, and I just have one question which you probably thought about already. Does it still work in the case that `rdim` is not specified, so it has the default value of `-1`?

siuwuncheung:
> I am approving since there are no example runs and nothing to verify, and I just have one question which you probably thought about already. Does it still work in the case that `rdim` is not specified, so it has the default value of `-1`?

Yes, if `rdim` is not specified, then the `overlap_count[curr_window]` will just be `windowOverlapSamples`, which is the same as original.

chldkdtn (approved):

pranav-vempati (approved):

PR #160, opened 2022-11-23
add destructors and virtual destructor to DMD base and derived classes

ebchin:
This PR adds destructors to `delete` allocated memory in `CAROM::DMD`, `CAROM::NonuniformDMD`, and `CAROM::AdaptiveDMD`.

### DIFF
--- a/lib/algo/DMD.cpp
+++ b/lib/algo/DMD.cpp
@@ -135,42 +135,15 @@ DMD::~DMD()
     {
         delete sampled_time;
     }
-    if (d_state_offset != nullptr)
-    {
-        delete d_state_offset;
-    }
-    if (d_basis != nullptr)
-    {
-        delete d_basis;
-    }
-    if (d_A_tilde != nullptr)
-    {
-        delete d_A_tilde;
-    }
-    if (d_phi_real != nullptr)
-    {
-        delete d_phi_real;
-    }
-    if (d_phi_imaginary != nullptr)
-    {
-        delete d_phi_imaginary;
-    }
-    if (d_phi_real_squared_inverse != nullptr)
-    {
-        delete d_phi_real_squared_inverse;
-    }
-    if (d_phi_imaginary_squared_inverse != nullptr)
-    {
-        delete d_phi_imaginary_squared_inverse;
-    }
-    if (d_projected_init_real != nullptr)
-    {
-        delete d_projected_init_real;
-    }
-    if (d_projected_init_imaginary != nullptr)
-    {
-        delete d_projected_init_imaginary;
-    }
+    delete d_state_offset;
+    delete d_basis;
+    delete d_A_tilde;
+    delete d_phi_real;
+    delete d_phi_imaginary;
+    delete d_phi_real_squared_inverse;
+    delete d_phi_imaginary_squared_inverse;
+    delete d_projected_init_real;
+    delete d_projected_init_imaginary;
 }
 
 void DMD::setOffset(Vector* offset_vector, int order)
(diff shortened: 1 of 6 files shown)


chldkdtn (approved):

pranav-vempati (approved):

lib/algo/DMD.cpp
```diff
@@ -125,6 +125,54 @@ DMD::DMD(std::vector<std::complex<double>> eigs, Matrix* phi_real,
...
+    }
+    if (d_basis != nullptr)
+    {
+        delete d_basis;
+    }
+    if (d_A_tilde != nullptr)
+    {
+        delete d_A_tilde;
+    }
+    if (d_phi_real != nullptr)
```
pranav-vempati, line 150: Nit: A nullptr in C++ is always falsy in boolean contexts, so e.g. if(d_phi_real) {...} should suffice here.

dylan-copeland:
@ebchin Did you run some DMD tests through valgrind to verify that all the memory is freed? Other memory leaks in libROM may be found, but it would be good at least to check the DMD classes.

lib/algo/DMD.cpp
dylan-copeland: Deleting `nullptr` is safe, but based on some internet discussions I'm not sure whether it is safe in more recent versions of C++. So it may be ok to remove the checks for `nullptr` and just delete. Otherwise, the suggestion by @pranav-vempati sounds like a good change to make here.

dylan-copeland: Also, note that DMD.h initializes to `NULL` rather than `nullptr`.

ebchin: Thanks for the comments, guys.  I couldn't find any discussion on whether `delete`ing a `nullptr` is not safe, so I opted to get rid of the `if` statements.  Would targeting a specific version of C++ in `compile.sh` (say C++11 or 14?) help avoid this problem @dylan-copeland?

lib/algo/NonuniformDMD.cpp
```diff
@@ -48,6 +48,14 @@ NonuniformDMD::NonuniformDMD(std::vector<std::complex<double>> eigs,
     setOffset(derivative_offset, 1);
 }
 
+NonuniformDMD::~NonuniformDMD()
+{
+    if (d_derivative_offset != nullptr)
```
dylan-copeland, line 53: One instance is remaining.

dylan-copeland (approved):

ebchin:
> @ebchin Did you run some DMD tests through valgrind to verify that all the memory is freed? Other memory leaks in libROM may be found, but it would be good at least to check the DMD classes.

Thanks for the suggestion, @dylan-copeland.  I ran `dg_advection` (serial) and valgrind found ~50 leaks in `initialize_matrix()` and `factorize()` in `scalapack_f_wrapper.f90`.  I also fixed a few leaks in `constructDMD()`.

ebchin:
I just verified all of the other DMD examples are still working and I ran valgrind on `heat_conduction`.  All of the leaks identified in `heat_conduction` within `CAROM::DMD` methods were limited to `factorize()` and `initialize_matrix()`.  If there are no other concerns, I will squash and merge later today.  Thanks!

chldkdtn:
> I just verified all of the other DMD examples are still working and I ran valgrind on `heat_conduction`. All of the leaks identified in `heat_conduction` within `CAROM::DMD` methods were limited to `factorize()` and `initialize_matrix()`. If there are no other concerns, I will squash and merge later today. Thanks!

Please go ahead and squash merge

PR #161, opened 2022-11-30
Fix pointwise sampling of L2 fields

dylan-copeland:
The examples in libROM do not seem to require or to be affected by this change. For some applications, the default averaging of points sampled on element boundaries in MFEM apparently has some issue that can result in negative values averaged from positive values. This PR sets the averaging type to `NONE`, avoiding this issue.


chldkdtn (approved):

chldkdtn:
@pranav-vempati @siuwuncheung let's review this quickly. This is only one line change PR.

siuwuncheung (approved):

PR #162, opened 2022-12-01
DMD time window variable basis dimensions

dylan-copeland:
This PR generalizes examples/dmd/parametric_tw_csv.cpp in multiple ways.

1. Each time window can now use its own DMD dimension, read from a CSV file, rather than prescribing a fixed dimension for all windows. This is necessary for robustness.
2. It can use HDF files partitioned in parallel by MPI rank. The sample runs in the comments, for a heat conduction example, were previously given with 8 MPI ranks. However, the problem is not actually distributed in parallel, so there is no advantage to parallelization. Since this heat conduction example is not yet set up to output partitioned data, this PR simply makes the sample runs serial. In general, parametric_tw_csv can be run on parallel distributed HDF files, with file names ending with `_R.hdf`, where `R` is the rank.
3. The naming of the HDF files is now refactored in the function `GetFilenameHDF`, and different naming conventions can be used by specifying `-hdfmode` in the command line. There are many places where the HDF file name is used, so refactoring allows for new naming convetions to be implemented in just one place, in `GetFilenameHDF`. For example, the default convention is
dmd_data/simS/fieldName_R.hdf
for simulation S and MPI rank R, whereas `-hdfmode 1` uses the convention
dmd_data/fieldNameS_R.hdf
4. Subsampling of snapshots is a new option, in both training and evaluation.
5. The predicted solution can optionally be output in an HDF file.


chldkdtn:
@dylan-copeland What is the status of this PR?

chldkdtn (approved):

examples/dmd/parametric_tw_csv.cpp
```diff
@@ -68,6 +68,25 @@
...
+        return string(data_dir) + "/" + sim_name + par_dir + "_"
+               + to_string(myid) + ".hdf";
+    default:
+        cout << string(data_dir) + "/" + sim_name + "/" + string(hdf_name) + "_"
+             + par_dir + "_" + to_string(myid) + ".hdf" << endl;
+        return string(data_dir) + "/" + sim_name + par_dir + "/" + string(hdf_name)
+               + "_" + to_string(myid) + ".hdf";
+    }
+}
+
```
siuwuncheung, line 89: I guess `mode` is related to two different file hierarchy assumptions? We may need to change the comments in Line 38-47.

siuwuncheung:
With the default value of `mode`, I guess it should not break any current capabilities. But since @debog and @ebchin both may use this script, I think it will be better for Debo and Eric to test if the changes in this PR won't break the current usage in their applications.

examples/dmd/parametric_tw_csv.cpp
dylan-copeland: The comments in Lines 38-47 describe the files for the CSV version, and Line 34 specifies that these comments are for the CSV version. The HDF version has multiple pieces of data within the same HDF file, so there is no need to have many files for all the data. The switch for the HDF version is the command-line option `-hdf`, so the default is CSV: `bool csvFormat = true;`

This `mode` on Line 76 is only an option within the HDF version. For example, field "energy" for parameter 3 and MPI rank 5 would have the HDF filename

mode 0: data/sim3/energy_5.hdf
mode 1: data/energy3_5.hdf

This is one of the changes described in the initial comments for this PR. It allows for either putting all the files in 1 directory (mode 1) or separate directories for each simulation (mode 0).

debog (approved):

siuwuncheung (approved):

examples/dmd/parametric_tw_csv.cpp
```diff
@@ -223,14 +259,21 @@ int main(int argc, char *argv[])
...
     int nelements = 0;
 
     if (csvFormat)
         db->getIntegerArray(string(data_dir) + "/dim.csv", &nelements, 1);
     else
     {
-        db->open(string(data_dir) + "/" + sim_name + "0/" + hdf_name, "r");
+        db->open(GetFilenameHDF(data_dir, sim_name, "0", hdf_name,
+                                myid, fileNameMode), "r");
         nelements = db->getDoubleArraySize("step0sol");
```
ebchin, line 277: This might be something for a future PR, but this should use `snap_pfx` and `snap_index_list` to be more general.  I've run into issues when I don't have a snapshot at 0.

ebchin (approved):
Looks good, thanks @dylan-copeland !

PR #165, opened 2022-12-09
Restore the snapshot filename format in CSV databases

siuwuncheung:
Several fixes are made to make the DMD capabilities work for general datasets. 

- Fix the temporal trimming (by getting the correct `snap_bound_size` from the file)
- Fix the snapshot filename restriction in CSV file. In general, the snapshot files may be named by a combination of prefix and suffix, `{snap_pfx}{snap_sfx}`, where `snap_pfx` is set as an option (default empty) and `snap_sfx` is read from the snapshot list. For example, in the heat conduction example, the option `-snap-pfx step` is added, and the first line of the list file is `0`. With the data format suffix `.csv`, the first snapshot filename is `step0.csv`. @dylan-copeland This changes will need you to add the corresponding option in your tests.
- Simplify the training phase of the `local_tw_csv` script.

### DIFF
--- a/examples/dmd/parametric_tw_csv.cpp
+++ b/examples/dmd/parametric_tw_csv.cpp
@@ -16,15 +16,15 @@
 // =================================================================================
 //
 // Parametric serial DMD command (for HDF version, append -hdf):
-//   mpirun -np 8 parametric_tw_csv -o hc_parametric_serial -rdim 16 -dtc 0.01 -offline
-//   mpirun -np 8 parametric_tw_csv -o hc_parametric_serial -rdim 16 -dtc 0.01 -online
+//   mpirun -np 8 parametric_tw_csv -o hc_parametric_serial -rdim 16 -dtc 0.01 -offline -snap-pfx step
+//   mpirun -np 8 parametric_tw_csv -o hc_parametric_serial -rdim 16 -dtc 0.01 -online -snap-pfx step
 //
 // Final-time prediction error (Last line in run/hc_parametric_serial/dmd_par5_prediction_error.csv):
 //   0.0012598331433506
 //
 // Parametric time windowing DMD command (for HDF version, append -hdf):
-//   mpirun -np 8 parametric_tw_csv -o hc_parametric_tw -nwinsamp 25 -dtc 0.01 -offline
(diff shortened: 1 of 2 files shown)


examples/dmd/parametric_tw_csv.cpp
```diff
@@ -16,15 +16,15 @@
...
+//   mpirun -np 8 parametric_tw_csv -o hc_parametric_serial -rdim 16 -dtc 0.01 -online -snap-pfx step
 //
 // Final-time prediction error (Last line in run/hc_parametric_serial/dmd_par5_prediction_error.csv):
 //   0.0012598331433506
 //
 // Parametric time windowing DMD command (for HDF version, append -hdf):
-//   mpirun -np 8 parametric_tw_csv -o hc_parametric_tw -nwinsamp 25 -dtc 0.01 -offline
-//   mpirun -np 8 parametric_tw_csv -o hc_parametric_tw -nwinsamp 25 -dtc 0.01 -online
+//   mpirun -np 8 parametric_tw_csv -o hc_parametric_tw -nwinsamp 25 -dtc 0.01 -offline -snap-pfx step
+//   mpirun -np 8 parametric_tw_csv -o hc_parametric_tw -nwinsamp 25 -dtc 0.01 -online -snap-pfx step
```
dylan-copeland, line 27: Should we make the default value `snap_pfx = "step"` so that this input argument can be skipped in the sample run?

dylan-copeland:
@siuwuncheung Reminder for `scripts/stylize.sh`. We need this as a CI check.

examples/dmd/parametric_tw_csv.cpp
siuwuncheung: The default option `snap_pfx = ""` works when there is no prefix, i.e. the full filename is written in the list. If we make the default `snap_pfx = "step"`, is there a way to make the option as empty `char` array when the executable is run? I try to write nothing after `-snap_pfx`, but it returns `Missing argument for the last option: -snap-pfx`.

dylan-copeland: Does `-snap_pfx ""` work?

siuwuncheung: It works. I changed the default value.

dylan-copeland (approved):

pranav-vempati (approved):

Issue #194, opened 2023-03-10
BasisGenerator not taking a trivial sample


dreamer2368:
`BasisGenerator::takeSample` does not take a trivial solution as sample, directed by the following if statement:

```
    // Check that u_in is not non-zero.
    Vector u_vec(u_in, getDim(), true);
    if (u_vec.norm() == 0.0) {
        return false;
    }
```

My intuition is that a trivial solution should be allowed to become a sample. Is there any special background or logic behind this?

siuwuncheung:
Theoretically, having the trivial vectors or not do not alter the singular vectors pattern. By omitting these trivial vectors, we at least do not lose anything. It might save some memory as well. But I believe the more important reason for omitting them is that if the incremental SVD  is used and the initial vector is zero (which can be the case, if we have zero initial condition), then the algorithm breaks down immediately in the normalization of the initial vector.

PR #200, opened 2023-04-11
Print out error message if trivial solution is taken.

dreamer2368:
Resolve the issue #194 .

`BasisGenerator` does not allow trivial samples, returning failure without any error message. Added a short error message to make it clear.

### DIFF
--- a/lib/linalg/BasisGenerator.cpp
+++ b/lib/linalg/BasisGenerator.cpp
@@ -132,7 +132,7 @@ BasisGenerator::takeSample(
     // Check that u_in is not non-zero.
     Vector u_vec(u_in, getDim(), true);
     if (u_vec.norm() == 0.0) {
-        printf("BasisGenerator::takeSample - trivial samples are not allowed!\n");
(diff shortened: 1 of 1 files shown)


lib/linalg/BasisGenerator.cpp
```diff
@@ -132,6 +132,7 @@ BasisGenerator::takeSample(
     // Check that u_in is not non-zero.
     Vector u_vec(u_in, getDim(), true);
     if (u_vec.norm() == 0.0) {
+        printf("BasisGenerator::takeSample - trivial samples are not allowed!\n");
```
dylan-copeland, line 135: This warning message is helpful, but it could be more informative. First, it may not be clear whether a serious error occurred, which would cause a failure. In this case, it is simply a warning that the sample was skipped. Maybe something like this would be clearer:

std::cout << "WARNING: BasisGenerator::takeSample skipped trivial sample at time " << time << std::endl;

dreamer2368: Revised the error message, reflecting this comment.

dylan-copeland (approved):

siuwuncheung (approved):

PR #207, opened 2023-05-05
Fix issue in doxygen

siuwuncheung:
It seems the issue is not in libROM, but a new release of [urllib3](https://github.com/urllib3/urllib3/releases/tag/2.0.0) is not compatible with the readthedocs docker image used to build the documentation. 

Following [this page](https://docs.readthedocs.io/en/stable/config-file/v2.html), it may be fixed by specifying the most recent image.

Update: the build #20414929 is successful.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,6 +7,8 @@ version: 2
 
 build:
   os: ubuntu-22.04
+  tools:
+    python: "3.11"
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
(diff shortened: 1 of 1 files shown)


siuwuncheung:
@ebchin I am curious if `build.os` was omitted intentionally when it was set up?

chldkdtn (approved):

ebchin:
> @ebchin I am curious if `build.os` was omitted intentionally when it was set up?

No, this was not intentional.  Thanks for the fix, @siuwuncheung!

ebchin (approved):

Issue #193, opened 2023-03-09
`test_RandomizedSVD` failure


dreamer2368:
The unit test `test_RandomizedSVD` fails. This is observed from both docker container [librom_env](https://hub.docker.com/repository/docker/dreamer2368/librom_env/general) and quartz.

This needs more investigations, but the issue seems to be a peculiar setup within the test.
For debug/test cases, `rand_mat` in `RandomizedSVD::computeSVD` is a matrix of 1.
As a result, the projected snapshot matrix `rand_proj` becomes 1-rank matrix.
The resulting `Q` matrix from qr factorization can have arbitrary orthonormal column basis except one column.
The result of SVD will depend on these arbitrary column basis of `Q`, although the answer is fixed to one case.

If this is the case, then the test `rand_mat` and the corresponding answer should be adjusted so that it may have some randomness (probably under the fixed random seed).

Following output is the result from `test_RandomizedSVD`:

```
[==========] Running 5 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 1 test from GoogleTestFramework
[ RUN      ] GoogleTestFramework.GoogleTestFrameworkFound
[       OK ] GoogleTestFramework.GoogleTestFrameworkFound (0 ms)
[----------] 1 test from GoogleTestFramework (0 ms total)

[----------] 4 tests from RandomizedSVDTest
[ RUN      ] RandomizedSVDTest.Test_RandomizedSVD
Distribution of sampler's A and U:
Local data: (  0.5377   -1.3077   -1.3499  
               1.8339   -0.4336    3.0349  
              -2.2588    0.3426    0.7254  
               0.8622    3.5784   -0.0631  
               0.3188    2.7694    0.7147  
             )
Process rank 0
=================
A->ctxt = 1
A->(m, n) = (5, 3)
A->(nprow, npcol) = (1, 1)
A->(pi, pj) = (0, 0)
A->(mb, nb) = (5, 5)
A->(mm, mn) = (5, 3)


Distribution of sampler's V:
Local data: ( -0.9093   -0.1798   -0.3752  
               0.4131   -0.2823   -0.8658  
              -0.0498    0.9423   -0.3310  
             )
Process rank 0
=================
A->ctxt = 2
A->(m, n) = (3, 3)
A->(nprow, npcol) = (1, 1)
A->(pi, pj) = (0, 0)
A->(mb, nb) = (5, 5)
A->(mm, mn) = (3, 3)


[...]

PR #211, opened 2023-05-14
Fix for `test_RandomizedSVD`

dreamer2368:
This PR resolves the issue #193 .

If `d_debug_algorithm` is true, then `RandomizedSVD` sets `rand_mat` to be identity matrix. This resolves the rank deficiency of the projected matrix `rand_proj`. Test answers are updated accordingly, verified separately from matlab.


siuwuncheung (approved):

lib/linalg/svd/RandomizedSVD.cpp
```diff
@@ -104,7 +104,7 @@ RandomizedSVD::computeSVD()
         rand_mat = new Matrix(snapshot_matrix->numColumns(), d_subspace_dim, false);
         for (int i = 0; i < snapshot_matrix->numColumns(); i++) {
             for (int j = 0; j < d_subspace_dim; j++) {
-                rand_mat->item(i, j) = 1;
+                rand_mat->item(i, j) = (i == j);
```
chldkdtn, line 107: need to change the text above, e.g., 

If debug mode is turned on, just set rand_mat as a matrix of 1s if i==j, zero otherwise.

chldkdtn (approved):

lib/linalg/svd/RandomizedSVD.cpp
dylan-copeland: Yes, the comment can just say it is set to the identity matrix in debug mode.

dylan-copeland (approved):

lib/linalg/svd/RandomizedSVD.cpp
siuwuncheung: It's a "truncated identity matrix" to be more accurate.

dreamer2368: changed the comments.

PR #216, opened 2023-05-31
Heat conduction with external source

siuwuncheung:
This PR adds a script `examples/dmd/heat_conduction_dmdc.cpp`. The initial condition is always assumed to be 1.0 globally in the spatial domain. The external source function is a single timed circular source/sink, which is parametrized by the circular radius `radius`, the source location `cx`, `cy`, the shut down time `t_end`, and the amplitude `amp`. 

The structure of this script is very much similar to the DMD for parametric heat conduction, though the parameterization is are different. The DMD components still remain in the script. Once DMDc is implemented, it should just require some minor changes (replacing the DMD routines by DMDc) on this script.

### DIFF
--- a/examples/dmd/heat_conduction_dmdc.cpp
+++ b/examples/dmd/heat_conduction_dmdc.cpp
@@ -130,6 +130,8 @@ public:
     ConductionOperator(ParFiniteElementSpace &f, double alpha, double kappa,
                        const Vector &u);
 
+    void GetSource(Vector& s) const;
+
     virtual void Mult(const Vector &u, Vector &du_dt) const;
     /** Solve the Backward-Euler equation: k = f(u + dt*k, t), for the unknown k.
         This is the only requirement for high-order SDIRK implicit integration.*/
--- a/examples/dmd/heat_conduction_dmdc.cpp
+++ b/examples/dmd/heat_conduction_dmdc.cpp
@@ -150,6 +150,7 @@ double radius = 0.5;
 double cx = 0.0;
 double cy = 0.0;
 double t_end = 0.1;
+double dt = 1.0e-2;
 double amp = 1.0;
 
 int main(int argc, char *argv[])
@@ -167,7 +168,6 @@ int main(int argc, char *argv[])
     int order = 2;
     int ode_solver_type = 3;
     double t_final = 0.5;
-    double dt = 1.0e-2;
     double alpha = 1.0e-2;
     double kappa = 0.5;
     double closest_rbf_val = 0.9;
@@ -1021,8 +1021,8 @@ double SourceFunction(const Vector &x, const double t)
     c.Elem(0) = cx;
     c.Elem(1) = cy;
     y -= c;
-    if (y.Norml2() < radius && t < t_end)
+    if (y.Norml2() < radius)
     {
-        return amp;
+        return amp / (1.0 + exp(0.5*(t-t_end)/dt));
     }
 }
--- a/examples/dmd/heat_conduction_dmdc.cpp
+++ b/examples/dmd/heat_conduction_dmdc.cpp
@@ -1,0 +1,1026 @@
+/******************************************************************************
+ *
+ * Copyright (c) 2013-2023, Lawrence Livermore National Security, LLC
+ * and other libROM project developers. See the top-level COPYRIGHT
+ * file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ *
+ *****************************************************************************/
+
+//                       libROM MFEM Example: Parametric_Heat_Conduction (adapted from ex16p.cpp)
+//
+// Compile with: make heat_conduction_dmdc
+//
+// =================================================================================
+//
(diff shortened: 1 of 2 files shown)


examples/dmd/heat_conduction_dmdc.cpp
```diff
@@ -0,0 +1,1028 @@
...
+    delete M;
+    delete K;
+}
+
+double InitialTemperature(const Vector &x)
+{
+    return 1.0;
+}
+
+double SourceFunction(const Vector &x, const double t)
```
chldkdtn, line 1017: We can make this source function continuously changing with respect to time variable, right? Not sure how DMDc will respond to this current step function in terms of time domain.

siuwuncheung: Yes, I can make the source function separable type, i.e. $f(x,t) = g(x)h(t)$.

siuwuncheung: Yes, I can make the source function become smooth.

chldkdtn (approved):

examples/dmd/heat_conduction_dmdc.cpp
```diff
@@ -0,0 +1,1028 @@
...
+                   "Enable or disable GLVis visualization.");
+    args.AddOption(&visit, "-visit", "--visit-datafiles", "-no-visit",
+                   "--no-visit-datafiles",
+                   "Save data files for VisIt (visit.llnl.gov) visualization.");
+    args.AddOption(&vis_steps, "-vs", "--visualization-steps",
+                   "Visualize every n-th timestep.");
+    args.AddOption(&rdim, "-rdim", "--rdim",
+                   "Reduced dimension for DMD.");
+    args.AddOption(&offline, "-offline", "--offline", "-no-offline", "--no-offline",
+                   "Enable or disable the offline phase.");
```
ebchin, line 237: It might be helpful to define offline vs online phase.  If I run online phase, it looks like it solves the ODE, but then loads snapshots later on?  I'm thinking online phase shouldn't solve the ODE.

examples/dmd/heat_conduction_dmdc.cpp
```diff
@@ -0,0 +1,1028 @@
...
+    for (int ti = 1; !last_step; ti++)
+    {
+        fom_timer.Start();
+
+        if (t + dt >= t_final - dt/2)
+        {
+            last_step = true;
+        }
+
+        ode_solver->Step(u, t, dt);
```
ebchin, line 593: This is what I'm referring to in the earlier comment; I don't think this needs to be done in the online phase?

examples/dmd/heat_conduction_dmdc.cpp
ebchin: I think one of the parameters (`cx`, `cy`, or `amp`) needs to be time dependent?

examples/dmd/heat_conduction_dmdc.cpp
```diff
@@ -0,0 +1,1028 @@
...
+    /// Update the diffusion BilinearForm K using the given true-dof vector `u`.
+    void SetParameters(const Vector &u);
+
+    virtual ~ConductionOperator();
+};
+
+double InitialTemperature(const Vector &x);
+double SourceFunction(const Vector &x, const double t);
+
+double radius = 0.5;
```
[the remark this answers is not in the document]
ebchin: On second thought, I think this is fine as is.

examples/dmd/heat_conduction_dmdc.cpp
```diff
@@ -0,0 +1,1028 @@
...
+//                       libROM MFEM Example: Parametric_Heat_Conduction (adapted from ex16p.cpp)
+//
+// Compile with: make heat_conduction_dmdc
+//
+// =================================================================================
+//
+// In these examples, the radius of the interface between different source functions, the
+// alpha coefficient, and two center location variables are modified.
+//
+// For Parametric DMD (ex. 1) (radius, interpolation):
```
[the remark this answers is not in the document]
siuwuncheung: These examples need to be replaced after DMDc is implemented by @michael-barrow-llnl. I'm leaving them for the file structure only.  (Also, for the other examples, I think the headers do not really exactly match the libROM website. Will fix the webpage when there is a chance.)

examples/dmd/heat_conduction_dmdc.cpp
siuwuncheung: The `amp` is really the amplitude at t = 0. The time dependence is in line 1026, where the amplitude of the source function decays smoothly.

examples/dmd/heat_conduction_dmdc.cpp
siuwuncheung: This is necessary for computing the reference solution at the testing parameter. Arguably, it is not part of the online phase in a ROM setting. But, as in other libROM examples on MFEM, I guess we want to minimize the number of phases to get the results.

ebchin (approved):

PR #227, opened 2023-07-11
IncrementalSVDFastUpdate with faster updates + bug fix in IncrementalSVD

swsuh28:
A new class`IncrementalSVDBrand : IncrementalSVD` has been added which exactly follows Brand's fast algorithm. It mostly follows `IncrementalSVDFastUpdate`, but has some modifications:

- Moved `computeBasis` from `buildIncrementalSVD` to `getSpatialBasis`. This is because we do not need to compute the basis while updating the SVD. We only need to compute the basis when we try to get its value.
- Added the re-orthogonalization of the projection error vector before being appended in `buildIncrementalSVD`. This guarantees the orthogonality of `d_U` down to the machine precision. (Gram-Schmidt alone does not guarantee the orthogonality down to the machine precision, but doing Gram-Schmidt twice does. See https://doi.org/10.1016/j.camwa.2005.08.009)
- Added the re-orthogonalization of `d_Up` and `d_W` in `addNewSample`. `d_U` is also being orthogonalized: although this will not be called since `d_U` is guaranteed to be orthonormal, it was not omitted just in case.

An additional test case has been added to `unit_test/test_IncrementalSVDBrand` to verify the revised version of the incremental SVD. It appends unit vectors to the incremental SVD object then checks if the computed basis and the singular values are either 0 or 1.

Finally, a bug in `IncrementalSVD::buildIncrementalSVD` has been fixed. Computing the norm of the projection error as
 
k = sqrt(u.u - 2.0*l.l + basisl.basisl)

should be avoided since this results in the catastrophic cancellation. It will lead to the incorrect estimate of the linear dependence of the new snapshot vector. Instead, one should compute it as

k = sqrt((u-basisl).(u-basisl)),

which remains double-precision.

### DIFF
--- a/lib/linalg/svd/IncrementalSVDFastUpdate.cpp
+++ b/lib/linalg/svd/IncrementalSVDFastUpdate.cpp
@@ -86,6 +86,18 @@ IncrementalSVDFastUpdate::~IncrementalSVDFastUpdate()
     }
 }
 
+const Matrix*
+IncrementalSVDFastUpdate::getSpatialBasis()
+{
+    // For fast update, we choose to compute the basis
+    // only when we try to get its value. Therefore,
+    // computeBasis() is called inside getSpatialBasis().
+    computeBasis(); // WARNING: this is costly
+
+    CAROM_ASSERT(d_basis != 0);
+    return d_basis;
+}
+
 void
 IncrementalSVDFastUpdate::buildInitialSVD(
     double* u,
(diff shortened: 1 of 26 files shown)


lib/linalg/svd/IncrementalSVDFastUpdate.cpp
```diff
@@ -136,9 +148,111 @@ IncrementalSVDFastUpdate::buildInitialSVD(
     d_num_samples = 1;
     d_num_rows_of_W = 1;
 
-    // Compute the basis vectors for this time interval.
-    computeBasis();
+}
+
+bool
+IncrementalSVDFastUpdate::buildIncrementalSVD(
```
dreamer2368, line 154: I believe this should be mathematically equivalent to the previous version, i.e. `IncrementalSVD::buildIncrementalSVD`. If so, can we add a unit test that checks the equivalence?

unit_tests/test_IncrementalSVD.cpp
```diff
@@ -138,6 +139,42 @@ TEST(IncrementalSVDSerialTest, Test_getSingularValues)
     }
 }
 
+TEST(IncrementalSVDSerialTest, Test_FastUpdate)
```
dreamer2368, line 142: This file is for the test of the base class. It may be better to move this inherited class test into a separate test file.

lib/linalg/svd/IncrementalSVDFastUpdate.cpp
```diff
@@ -86,6 +86,18 @@ IncrementalSVDFastUpdate::~IncrementalSVDFastUpdate()
     }
 }
 
+const Matrix*
+IncrementalSVDFastUpdate::getSpatialBasis()
+{
+    // For fast update, we choose to compute the basis
+    // only when we try to get its value. Therefore,
+    // computeBasis() is called inside getSpatialBasis().
+    computeBasis(); // WARNING: this is costly
```
dreamer2368, line 95: `computeBasis` doe not only compute `d_basis`, but also `d_basis_right` and `d_S`. If we defer `computeBasis` until we call basis functions, then we need to ensure `getTemporalBasis` and `getSingularValues` also call `computeBasis`. At the same time, it should be also ensured that `computeBasis` is not called repeatedly, when calling these `get--` functions.

unit_tests/test_IncrementalSVD.cpp
```diff
@@ -138,6 +139,42 @@ TEST(IncrementalSVDSerialTest, Test_getSingularValues)
...
+            .setIncrementalSVD(1e-1, 1e-1, 1e-1, 1e-1, fast_update);
+
+    CAROM::BasisGenerator bg(
+        incremental_svd_options,
+	true,
+        "irrelevant.txt");
+
+    for (int i = 0; i < dim; i++)
+    {
+	// Unit vector
```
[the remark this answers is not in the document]
swsuh28: Adopted a test case from `test_RandomizedSVD` to the new unit test file `test_IncrementalSVDBrand.cpp`. Both MPI and non-MPI versions were added to `action.yml`.

lib/linalg/svd/IncrementalSVDFastUpdate.cpp
swsuh28: `IncrementalSVDFastUpdate` has been restored to the original version, and all the new features have been moved to a new class `IncrementalSVDBrand`. `comupteBasis` now calls `updateSpatialBasis` and `updateTemporalBasis` to compute basis matrices. These `update--` functions are also individually called in each `get--` functions. For example, `getTemporalBasis` calls only `updateTemporalBasis` before returning `d_basis_right` and does not update `d_basis`, which could be costly and unnecessary.

dreamer2368 (approved):

chldkdtn (approved):
Very nice!

siuwuncheung (approved):

PR #230, opened 2023-07-16
ScaLAPACK dependency installation

dreamer2368:
This is in support of [pylibROM installation](https://github.com/michael-barrow-llnl/pylibROM/pull/12) using `pip` on quartz.

Current `libROM` compilation using `mkl` module in `quartz` has an issue, as the shared object of `libROM` cannot resolve the relative path of `mkl`, as the shared object is packed into the pip installation procedure. (For a detail, see the issue from [pybind/cmake_example](https://github.com/pybind/cmake_example/issues/11)).

- `scripts/setup.sh` is adjusted so that it can manually install a static `libscalapack.a` library. There are two options to specify manual installation,
 1. specify `-s` flag for `scripts/compile.sh`
 2. pass `true` as an argument after `scripts/setup.sh`
- Required source code and input file is stored in `dependencies`.
- `lib/CMakeLists.txt` provides `HINTS` to `find_package(ScaLAPACK REQUIRED)`, pointing to `dependencies/scalapack-2.2.0`.

A miscellaneous addition:
- `cmake/toolchains/simple.cmake`: it simply takes the path to `mfem`, `hypre`, `parmetis` and `metis` from the environment variables (which need to be pre-set).

### DIFF
--- a/cmake/toolchains/simple.cmake
+++ b/cmake/toolchains/simple.cmake
@@ -1,0 +1,9 @@
+set(CMAKE_C_COMPILER mpicc)
+set(CMAKE_CXX_COMPILER mpicxx)
+set(CMAKE_Fortran_COMPILER mpif90)
+
+# Use environment variables.
+set(MFEM_DIR $ENV{MFEM_DIR})
+set(HYPRE_DIR $ENV{HYPRE_DIR})
+set(PARMETIS_DIR $ENV{PARMETIS_DIR})
+set(METIS_DIR $ENV{METIS_DIR})
--- a/dependencies/SLmake.inc
+++ b/dependencies/SLmake.inc
@@ -26,6 +26,11 @@ CDEFS         = -DAdd_
 #  The fortran and C compilers, loaders, and their flags
 #
 
+# LIBROM instruction: copy this file to scalapack-2.2.0 directory and type make.
+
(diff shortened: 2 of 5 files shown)


chldkdtn:
@dreamer2368 Thank you for looking into this. Wonderful job. By the way, it seems that libROM still extract scalapack tar files in dependencies and install it even though it has been done in the past. Can you modify it so that if the scalapack is already installed, then it skips the extraction and re-installing? Perhaps, for pylibrom, we may want to make an option of forcing scalapack extraction and installation to avoid INTEL-MKL errors. What do you think?

dreamer2368:
> @dreamer2368 Thank you for looking into this. Wonderful job. By the way, it seems that libROM still extract scalapack tar files in dependencies and install it even though it has been done in the past. Can you modify it so that if the scalapack is already installed, then it skips the extraction and re-installing? Perhaps, for pylibrom, we may want to make an option of forcing scalapack extraction and installation to avoid INTEL-MKL errors. What do you think?

The `scripts/setup.sh` now further checks if the `libscalapack.a` exists in the dependencies.

chldkdtn (approved):

michael-barrow-llnl (approved):
I'm quite happy this has been done, I did have to resolve the ScaLAPACK dependency on non-LC machines with some modifications to the CMake which use open versions. Will discuss if this (non-LC issue) does not go away and if it turns out that that the CMake modification was desirable. Approving for now.

PR #235, opened 2023-08-15
StaticSVD hot fix for transposed SVD operation.

dreamer2368:
When the sample size is larger than the system dimension, transposed SVD operation is executed in `StaticSVD`. In this case, `ncolumns` must be less than or equal to the system dimension. Previously, this is not automatically enforced, unless `d_max_basis_dimension` is manually specified. This causes the scalapack `Bad_Submatrix` error in `StaticSVD::computeSVD`.

Now this is fixed by enforcing `ncolumns` to be limited by the system dimension, when the transposed SVD is executed.

### DIFF
--- a/unit_tests/test_IncrementalSVDBrand.cpp
+++ b/unit_tests/test_IncrementalSVDBrand.cpp
@@ -72,9 +72,9 @@ TEST(IncrementalSVDBrandTest, Test_IncrementalSVDBrand)
 
     double* basis_right_true_ans = new double[9] {
         -1.78651649346571794741E-01,     5.44387957786310106023E-01,      -8.19588518467042281834E-01,
-        -9.49719639253861602768E-01,     -3.13100149275943651084E-01,     -9.50441422536040881122E-04,
-        -2.57130696341890396805E-01,     7.78209514167382598870E-01,      5.72951792961765460355E-01
-    };
+            -9.49719639253861602768E-01,     -3.13100149275943651084E-01,     -9.50441422536040881122E-04,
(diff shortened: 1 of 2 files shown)


unit_tests/test_IncrementalSVDBrand.cpp
```diff
@@ -72,9 +72,9 @@ TEST(IncrementalSVDBrandTest, Test_IncrementalSVDBrand)
 
     double* basis_right_true_ans = new double[9] {
         -1.78651649346571794741E-01,     5.44387957786310106023E-01,      -8.19588518467042281834E-01,
-        -9.49719639253861602768E-01,     -3.13100149275943651084E-01,     -9.50441422536040881122E-04,
```
dylan-copeland, line 75: This file is changed by `astyle` 2.05, but it is unchanged by 3.1. We do not have an official version of `astyle` in libROM, but I suppose it makes sense to progress to 3.1 and not revert changes back to 2.05.

dreamer2368: I guess you're meaning that the current PR is following astyle 3.1 convention? (Which should be, since it is enforced by ci workflow `code-style`) By the fact that the ci workflow enforces style with v3.1, we might as well say astyle v.3.1 is our official stylizer?

dylan-copeland: Yes, actually I was confused about whether this was using 3.1 or 2.05. I think we should use 3.1 or 3.2 and make it official.

dylan-copeland (approved):

chldkdtn (approved):

PR #237, opened 2023-09-14
Minor fix in mfem utility function + Hypre version change

dreamer2368:
This PR is in support of pylibROM mfem binding [PR](https://github.com/LLNL/pylibROM/pull/14).

When using mfem functions from python side, a strange seg-fault error happens when `HypreParMatrix::GetNumRows()` is used. This does not happen if the same function is used from c++ side.

Fortunately, there is `HypreParMatrix::NumRows()` which does the same job and does not cause a seg-fault issue on python end. This PR simply replaces `GetNumRows()` with `NumRows()`. 

### Update
The seg-fault error in using Hypre functions on python end, turns out to be because of the version incompatibility between libROM and pymfem.

[PyMFEM](https://github.com/mfem/PyMFEM) uses hypre v2.28.0, while libROM has been using v2.20.0. And when pylibrom accesses to some hypre functions compiled from pymfem (or the other way around), it causes the segmentation fault error.

This error is resolved when the hypre in libROM is updated to v2.28.0. So now the `setup.sh` and `Dockerfile` reflect this version change.

### Another update

Further segmentation fault error is observed, and this was due to the version mismatch of  c++ `mfem` between pylibROM and PyMFEM. To support their interface, we match the mfem version to that from PyMFEM.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -93,8 +93,8 @@ WORKDIR $LIB_DIR
 # install mfem
 RUN git clone https://github.com/mfem/mfem.git mfem_parallel
 WORKDIR ./mfem_parallel
-# Latest verified commit on May 2, 2023
-RUN git checkout e5231334e6a8175b4f404b877f590b73dee2dedc
+# v4.5.2-dev commit. This is the mfem version used by PyMFEM v4.5.2.0.
+RUN git checkout 00b2a0705f647e17a1d4ffcb289adca503f28d42
 #RUN tar -zxvf mfem-4.5.tar.gz
 #RUN mv mfem-4.5 mfem_parallel
 #RUN git pull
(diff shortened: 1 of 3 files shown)


dylan-copeland:
The difference between these functions is that `GetNumRows` calls a hypre function, whereas `NumRows` simply returns `height` in the base class `Operator`. I guess the hypre call has some complication in the python version, and maybe that needs to be resolved in pymfem. It is not clear from the header files whether `GetNumRows` and `NumRows` should always return the same number, but it looks like the `HypreParMatrix` constructors set `height` to `GetNumRows`, so I think there should not be any difference. At least the regression tests passed.

dylan-copeland (approved):

dylan-copeland (approved):

siuwuncheung (approved):

PR #239, opened 2023-09-23
Changes in Gram-Schmidt functions

cval26:
This PR includes two main changes. 

1. Fixes issues in the current `orthogonalize` method of the `Matrix` class.
2. Adds an incremental Gram-Schmidt function `orthogonalize_last`, where only the last column of the input matrix is orthonormalized, assuming the previous columns are already orthonormal. 

The above changes are motivated by applications in the Laghos hydrodynamics code.

### DIFF
--- a/lib/linalg/Matrix.cpp
+++ b/lib/linalg/Matrix.cpp
@@ -1818,7 +1818,7 @@ Matrix::orthogonalize()
         // Normalize the column.
         double norm = 0.0;
         tmp = 0.0;
-        
+
         for (int i = 0; i < d_num_rows; ++i)
             tmp += item(i, work) * item(i, work);
 
@@ -1844,7 +1844,7 @@ Matrix::orthogonalize_last(int ncols)
 {
     if (ncols == -1) ncols = d_num_cols;
     CAROM_ASSERT((ncols > 0) && (ncols <= d_num_cols));
-    
+
     const int last_col = ncols - 1; // index of column to be orthonormalized
     double tmp;
 
--- a/lib/linalg/Matrix.cpp
+++ b/lib/linalg/Matrix.cpp
@@ -1788,28 +1788,22 @@ const
 }
 
 void
-Matrix::orthogonalize()
+Matrix::orthogonalize(double zero_tol)
 {
     for (int work = 0; work < d_num_cols; ++work)
     {
         // Orthogonalize the column.
-        double tmp;
         for (int col = 0; col < work; ++col)
         {
             double factor = 0.0;
-            tmp = 0.0;
 
             for (int i = 0; i < d_num_rows; ++i)
-                tmp += item(i, col) * item(i, work);
+                factor += item(i, col) * item(i, work);
 
             if (d_distributed && d_num_procs > 1)
             {
-                MPI_Allreduce(&tmp, &factor, 1, MPI_DOUBLE, MPI_SUM,
-                              MPI_COMM_WORLD);
-            }
-            else
-            {
-                factor = tmp;
+                CAROM_VERIFY( MPI_Allreduce(MPI_IN_PLACE, &factor, 1,
+                                            MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD) == MPI_SUCCESS );
             }
             for (int i = 0; i < d_num_rows; ++i)
                 item(i, work) -= factor * item(i, col);
@@ -1817,20 +1811,16 @@ Matrix::orthogonalize()
 
         // Normalize the column.
         double norm = 0.0;
-        tmp = 0.0;
 
         for (int i = 0; i < d_num_rows; ++i)
-            tmp += item(i, work) * item(i, work);
+            norm += item(i, work) * item(i, work);
 
         if (d_distributed && d_num_procs > 1)
         {
-            MPI_Allreduce(&tmp, &norm, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+            CAROM_VERIFY( MPI_Allreduce(MPI_IN_PLACE, &norm, 1, MPI_DOUBLE,
+                                        MPI_SUM, MPI_COMM_WORLD) == MPI_SUCCESS );
         }
-        else
-        {
-            norm = tmp;
-        }
-        if (norm > 1.0e-15)
+        if (norm > zero_tol)
         {
             norm = 1.0 / sqrt(norm);
             for (int i = 0; i < d_num_rows; ++i)
@@ -1840,31 +1830,25 @@ Matrix::orthogonalize()
 }
 
 void
-Matrix::orthogonalize_last(int ncols)
+Matrix::orthogonalize_last(int ncols, double zero_tol)
 {
     if (ncols == -1) ncols = d_num_cols;
-    CAROM_ASSERT((ncols > 0) && (ncols <= d_num_cols));
+    CAROM_VERIFY((ncols > 0) && (ncols <= d_num_cols));
 
     const int last_col = ncols - 1; // index of column to be orthonormalized
-    double tmp;
 
     // Orthogonalize the column.
     for (int col = 0; col < last_col; ++col)
     {
         double factor = 0.0;
-        tmp = 0.0;
 
         for (int i = 0; i < d_num_rows; ++i)
-            tmp += item(i, col) * item(i, last_col);
+            factor += item(i, col) * item(i, last_col);
 
         if (d_distributed && d_num_procs > 1)
         {
-            MPI_Allreduce(&tmp, &factor, 1, MPI_DOUBLE, MPI_SUM,
-                          MPI_COMM_WORLD);
-        }
-        else
-        {
-            factor = tmp;
+            CAROM_VERIFY( MPI_Allreduce(MPI_IN_PLACE, &factor, 1, MPI_DOUBLE,
+                                        MPI_SUM, MPI_COMM_WORLD) == MPI_SUCCESS );
         }
         for (int i = 0; i < d_num_rows; ++i)
             item(i, last_col) -= factor * item(i, col);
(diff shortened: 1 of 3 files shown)


lib/linalg/Matrix.cpp
```diff
@@ -1790,46 +1790,95 @@ const
 void
 Matrix::orthogonalize()
 {
-    for (int work = 1; work < d_num_cols; ++work) {
+    for (int work = 0; work < d_num_cols; ++work)
```
cval26, line 1793: This is the real change in this function; ensuring that the first column is normalized. The rest is changes to make the code easier to read; no functionality changes.

lib/linalg/Matrix.h
```diff
@@ -946,6 +946,19 @@ class Matrix
...
+	 * columns are already orthonormal.
+	 * 
+	 * By default, the function considers the whole matrix.
+	 * If input parameter ncols < d_num_cols, then a subset of the matrix
+	 * is considered. 
+	 * This allows one to reorthonormalize the matrix every time a new column
+	 * is added, assuming the previous columns have remained unchanged.
+     */
+	void
+	orthogonalize_last(const int ncols = d_num_cols);
```
cval26, line 960: @dylan-copeland 
In this function, I want to set the default value for the input argument to be the number of columns of the matrix calling this method. In this way, if no input argument is given, the last column is orthonormalized. If a value smaller than `d_num_cols` is given, then the function considers a subset of the full matrix, allowing the user to reorthonormalize every time a new column is added to a matrix with a larger allocated memory size. 

Given that it is a method of class `Matrix`, my understanding is that the function should have access to the `d_num_cols` variable. But when trying to compile I get the following error: "a nonstatic member reference must be relative to a specific object". The same happens if I try to use the `numColumns()` method instead.

dylan-copeland: I don't think a class member (here, `d_num_cols`) can be used as the default value for a function input argument. Instead, you could set to `d_num_cols` inside the function. For example, the default could be something meaningless, like `const int ncols = -1`, which serves as a flag inside the function: `if (ncols < 0) ncols = d_num_cols;`

dylan-copeland:
The style script needs to be run.

cval26:
@dylan-copeland
I just did. Sorry about that; didn't run it right the first time.

cval26:
The tests I added in the `Matrix` test suite fail in the CI build, even though they run successfully in my local build on LC. I'll have to check why and get back to you.

lib/linalg/Matrix.cpp
```diff
@@ -1790,48 +1790,109 @@ const
...
-                MPI_Allreduce(&tmp,
-                              &factor,
-                              1,
-                              MPI_DOUBLE,
-                              MPI_SUM,
+
+            for (int i = 0; i < d_num_rows; ++i)
+                tmp += item(i, col) * item(i, work);
+
+            if (d_num_procs > 1)
```
dylan-copeland, line 1805:
```suggestion
            if (d_distributed && d_num_procs > 1)
```

lib/linalg/Matrix.cpp
```diff
@@ -1790,48 +1790,109 @@ const
...
         tmp = 0.0;
-        for (int i = 0; i < d_num_rows; ++i) {
-            tmp += item(i, work)*item(i, work);
-        }
-        if (d_num_procs > 1) {
+
+        for (int i = 0; i < d_num_rows; ++i)
+            tmp += item(i, work) * item(i, work);
+
+        if (d_num_procs > 1)
```
dylan-copeland, line 1825:
```suggestion
        if (d_distributed && d_num_procs > 1)
```

lib/linalg/Matrix.cpp
```diff
@@ -1790,48 +1790,109 @@ const
...
+    // Orthogonalize the column.
+    for (int col = 0; col < last_col; ++col)
+    {
+        double factor = 0.0;
+        tmp = 0.0;
+
+        for (int i = 0; i < d_num_rows; ++i)
+            tmp += item(i, col) * item(i, last_col);
+
+        if (d_num_procs > 1)
```
dylan-copeland, line 1860:
```suggestion
        if (d_distributed && d_num_procs > 1)
```

lib/linalg/Matrix.cpp
```diff
@@ -1790,48 +1790,109 @@ const
...
+    }
+
+    // Normalize the column.
+    double norm = 0.0;
+    tmp = 0.0;
+
+    for (int i = 0; i < d_num_rows; ++i)
+        tmp += item(i, last_col) * item(i, last_col);
+
+    if (d_num_procs > 1)
```
dylan-copeland, line 1880:
```suggestion
    if (d_distributed && d_num_procs > 1)
```

lib/linalg/Matrix.cpp
```diff
@@ -1790,45 +1790,106 @@ const
...
+    // Orthogonalize the column.
+    for (int col = 0; col < last_col; ++col)
+    {
+        double factor = 0.0;
+        tmp = 0.0;
+
+        for (int i = 0; i < d_num_rows; ++i)
+            tmp += item(i, col) * item(i, last_col);
+
+        if (d_distributed && d_num_procs > 1)
```
dylan-copeland, line 1860: This can be simplified a bit by reducing in place, eliminating `tmp`:

if (d_distributed && d_num_procs > 1) {
   CAROM_VERIFY(MPI_Allreduce(MPI_IN_PLACE, &factor, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD) == MPI_SUCCESS);
    }

lib/linalg/Matrix.cpp
```diff
@@ -1790,45 +1790,106 @@ const
...
+                item(i, work) *= norm;
+        }
+    }
+}
+
+void
+Matrix::orthogonalize_last(int ncols)
+{
+    if (ncols == -1) ncols = d_num_cols;
+    CAROM_ASSERT((ncols > 0) && (ncols <= d_num_cols));
```
dylan-copeland, line 1846: `CAROM_ASSERT` is checked only in debug builds, whereas `CAROM_VERIFY` is always checked. The former is preferable when the check may affect performance, and the latter is preferable when there is no impact on performance. Since this check is not inside the computations, I would prefer `CAROM_VERIFY` here.

lib/linalg/Matrix.cpp
```diff
@@ -1790,45 +1790,106 @@ const
...
+
+    const int last_col = ncols - 1; // index of column to be orthonormalized
```
dylan-copeland, line 1848: This means `ncols` is 1-based, but we always use 0-based indices in libROM.

lib/linalg/Matrix.h
```diff
@@ -946,6 +946,19 @@ class Matrix
     void
     orthogonalize();
 
+    /**
+     * @brief Orthonormalizes the matrix's last column, assuming the previous
+     * columns are already orthonormal.
+     *
+     * By default, the function considers the whole matrix.
```
dylan-copeland, line 953: To be a bit more precise, you could say `By default (ncols == -1)`.

lib/linalg/Matrix.cpp
```diff
@@ -1790,45 +1790,106 @@ const
...
         }
-        else {
+        else
+        {
             norm = tmp;
         }
-        norm = sqrt(norm);
-        for (int i = 0; i < d_num_rows; ++i) {
-            item(i, work) /= norm;
+        if (norm > 1.0e-15)
```
dylan-copeland, line 1833: Should there be a `CAROM_VERIFY(norm > 1.0e-15);`? If the column becomes 0, shouldn't that be a failure?

chldkdtn: @cval26 are you going to address this?

cval26: Yes, I'm going to make the required changes this week.

lib/linalg/Matrix.cpp
cval26: Done.

lib/linalg/Matrix.cpp
cval26: Done.

lib/linalg/Matrix.cpp
cval26: The function takes as input the total *number* of columns of the matrix, not their indices. The index that corresponds to the last column is calculated internally. 

I find it easier for the user to have to provide the number of columns present in the matrix, rather than the index that corresponds to the last column.

lib/linalg/Matrix.cpp
cval26: I don't think it should be a failure if a column is identically zero. For instance, when using Laghos EQP with the `load` offset option, the first NNLS constraints column can be zero, because the first snapshot vectors are identically zero.

cval26: Added a private double `zero_tol` variable and a public method for setting its value. Default value is `1e-15`.

lib/linalg/Matrix.h
cval26: Done.

lib/linalg/Matrix.cpp
cval26: I don't think it should be a failure if a column becomes zero. This means that the input matrix had linearly dependent columns, so the method has reduced one column to essentially zero.

lib/linalg/Matrix.h
```diff
@@ -941,11 +941,24 @@ class Matrix
...
      */
     void
     orthogonalize();
 
+    /**
+     * @brief Orthonormalizes the matrix's last column, assuming the previous
+     * columns are already orthonormal.
+     *
+     * By default, ncols == -1, and the function considers the whole matrix.
+     * If ncols /= -1 and ncols < d_num_cols, then a subset of the matrix
```
dylan-copeland, line 954: If ncols != -1

lib/linalg/Matrix.h
```diff
@@ -1260,6 +1279,13 @@ class Matrix
      * If d_owns_data is false, then the object may not reallocate d_mat.
      */
     bool d_owns_data;
+
+    /**
+     * @brief If the absolute value of a double variable is below this
+     * tolerance, it is assumed to be zero.
+     * It is used to decide if we can safely divide by a value.
```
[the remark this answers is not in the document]
dylan-copeland: On second thought, since this number is only used in the functions `orthogonalize` and `orthogonalize_last`, maybe it is better to make this an input parameter with default value, e.g.
`orthogonalize(double zero_tol=1.0e-15)`

dylan-copeland:
`git pull origin master` will resolve the conflicts and should make the regression tests pass (currently they are using wrong command line flags).

lib/linalg/Matrix.h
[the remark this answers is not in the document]
cval26: Done.

lib/linalg/Matrix.h
cval26: Done.

siuwuncheung (approved):

dylan-copeland (approved):

PR #247, opened 2023-11-09
Fix memory issue in VectorInterpolator

ckendrick:
This updates the destructor of `VectorInterpolator` to only free the memory it allocates, similar to how the `MatrixInterpolator` was changed in #179. This fixes a memory issue that appeared in the DG local advection example and python bindings in https://github.com/LLNL/pylibROM/pull/22.

### DIFF
--- a/lib/algo/manifold_interp/VectorInterpolator.cpp
+++ b/lib/algo/manifold_interp/VectorInterpolator.cpp
@@ -61,19 +61,29 @@ VectorInterpolator::VectorInterpolator(std::vector<Vector*> parameter_points,
         if (i == d_ref_point)
         {
             d_rotated_reduced_vectors.push_back(reduced_vectors[i]);
+            d_rotated_reduced_vectors_owned.push_back(false);
         }
         else
         {
             Vector* Q_tA = rotation_matrices[i]->transposeMult(reduced_vectors[i]);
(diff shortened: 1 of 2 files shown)


dylan-copeland (approved):

lib/algo/manifold_interp/VectorInterpolator.cpp
```diff
@@ -61,19 +61,29 @@ VectorInterpolator::VectorInterpolator(std::vector<Vector*> parameter_points,
...
-    for (auto v : d_rotated_reduced_vectors)
-        delete v;
+    CAROM_VERIFY(d_rotated_reduced_vectors.size() ==
+                 d_rotated_reduced_vectors_owned.size());
+
+    for (int i = 0; i < d_rotated_reduced_vectors.size(); i++)
+    {
+        if (d_rotated_reduced_vectors_owned[i])
+        {
+            delete d_rotated_reduced_vectors[i];
```
siuwuncheung, line 84: I guess it is fine if we make sure it is never re-used again, but should `d_rotated_reduced_vectors_owned[i]` be made `false` just in case?

dylan-copeland: I think it doesn't matter, since it is in the destructor, and the flag array will be destroyed as well.

siuwuncheung (approved):

PR #255, opened 2023-12-19
Double pass Gram-Schmidt methods

cval26:
Adds the option of reorthogonalizing each column of the input matrix to the `orthogonalize()` and `orthogonalize_last()` methods of the `Matrix` class. 

The option is added as an input argument which is false by default, thereby not affecting the methods' default behavior. The only change to the existing code is that the orthogonalizing part of the methods is run twice for each column, if the double pass option is true. 

Both the classical and modified Gram-Schmidt numerical algorithms are prone to loss of orthogonality, which can result in output matrices that deviate significantly from being orthonormal. Orthogonalizing each column twice has been shown to mitigate this issue, see [Giraud, Langou & Rozloznik (2005)](https://doi.org/10.1016/j.camwa.2005.08.009).

Adding the double-pass option is motivated by its use in LaghosROM, where it's been found to successfully address the loss of orthogonality issue, improving accuracy.

### DIFF
--- a/lib/linalg/Matrix.cpp
+++ b/lib/linalg/Matrix.cpp
@@ -1830,28 +1830,32 @@ Matrix::orthogonalize(double zero_tol)
 }
 
 void
-Matrix::orthogonalize_last(int ncols, double zero_tol)
+Matrix::orthogonalize_last(int ncols, bool double_pass, double zero_tol)
 {
     if (ncols == -1) ncols = d_num_cols;
     CAROM_VERIFY((ncols > 0) && (ncols <= d_num_cols));
 
     const int last_col = ncols - 1; // index of column to be orthonormalized
 
-    // Orthogonalize the column.
-    for (int col = 0; col < last_col; ++col)
+    // Orthogonalize the column (twice if double_pass == true).
+    for (int k = 0; k < 2; k++)
     {
-        double factor = 0.0;
-
-        for (int i = 0; i < d_num_rows; ++i)
-            factor += item(i, col) * item(i, last_col);
-
-        if (d_distributed && d_num_procs > 1)
+        for (int col = 0; col < last_col; ++col)
         {
-            CAROM_VERIFY( MPI_Allreduce(MPI_IN_PLACE, &factor, 1, MPI_DOUBLE,
-                                        MPI_SUM, MPI_COMM_WORLD) == MPI_SUCCESS );
+            double factor = 0.0;
+
+            for (int i = 0; i < d_num_rows; ++i)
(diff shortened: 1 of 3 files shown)


cval26:
The git diff tracker seems to be confused by the changes made in the iteration loops. By viewing the files in split mode, you can see that the only change is adding a loop over the orthogonalizing part of the code and an exit condition based on the `double_pass` flag. 

This should make reviewing much more straightforward.

lib/linalg/Matrix.cpp
```diff
@@ -1788,25 +1788,31 @@ const
...
 void
-Matrix::orthogonalize(double zero_tol)
+Matrix::orthogonalize(bool double_pass, double zero_tol)
 {
     for (int work = 0; work < d_num_cols; ++work)
     {
-        // Orthogonalize the column.
-        for (int col = 0; col < work; ++col)
+        // Orthogonalize the column (twice if double_pass == true).
+        for (int k = 0; k < 2; k++)
```
ckendrick, line 1796: This is probably a bit picky, but to try to reduce the number of if-statements inside the outer `work` loop, we could define the number of passes at the top of the function outside of the loop:
`const int num_pass = double_pass ? 2 : 1;`
Then the `k` loop could then be re-written as `for (int k = 0; k < num_pass; k++)` and the `if (!double_pass) break;` line can be removed.

cval26:
@dylan-copeland @siuwuncheung Your review of this PR is still pending.

lib/linalg/Matrix.h
```diff
@@ -943,13 +943,19 @@ class Matrix
...
+     *
      * If the norm of a matrix column is below the value of zero_tol then it
      * is considered to be zero, and we do not divide by it.
      * Therefore, that column is considered to be zero and is not normalized.
-     * By default, zero_tol = 1.0e-15.
+     * By default, zero_tol == 1.0e-15.
      */
     void
-    orthogonalize(double zero_tol = 1.0e-15);
+    orthogonalize(bool double_pass = false, double zero_tol = 1.0e-15);
```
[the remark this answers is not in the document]
cval26: Yes, this is why I placed `double pass` first. I added the `zero_tol` argument very recently and know that no code is explicitly setting a value for this parameter, so moving it to second place is in fact not a breaking change.

lib/linalg/Matrix.cpp
cval26: Done.

ckendrick (approved):

dylan-copeland (approved):

PR #257, opened 2023-12-21
Add toolchain for Lassen

ckendrick:
This modifies some of the setup and compile scripts and adds a CMake toolchain file to make libROM compile on the IBM Power9 machines (such as Lassen).

Currently this is toolchain is only for the GNU 12 compilers. A toolchain using the IBM XL compiler will need some more work and will be added in a separate PR.

**To Test:**
Due to some of the system module defaults and how the libROM dependencies are compiled, the following modules will need to be loaded before using `compile.sh` with the toolchain.

```
module load cmake/3.23.1
module load gcc/12.2.1
module load lapack/3.11.0-gcc-11.2.1
export CFLAGS="-mcpu=powerpc64le -mtune=powerpc64le"
export CXXFLAGS="-mcpu=powerpc64le -mtune=powerpc64le"
```

**Note:** if libROM has not been cloned from scratch, be sure to remove all folders in `dependencies/` to ensure they get recompiled for the correct architecture.

Then the toolchain can be used to compile libROM (with MFEM and GSLIB):
```
./scripts/compile.sh -d -g -m -t cmake/toolchains/gnu12-rhel_7_ppc64le_ib_p9-dev.cmake
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,12 @@ Compilation options:
 - -t: Use your own cmake/toolchain
 - -u: Update all of libROM's dependencies.
 
+## Compiling on LC Machines
+
+libROM provides several CMake toolchains which can be used to compile on LLNL LC machines. 
+For more information on installing and using libROM on specific LC machines, 
+refer to [this wiki page](https://github.com/LLNL/libROM/wiki/Compiling-on-LC-Machines).
+
 # Installing via Spack
 
 There is a Spack package for libROM; however, the version it installs
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ LLNL-CODE-766763
 
 # Authors
 - Robert W. Anderson (LLNL)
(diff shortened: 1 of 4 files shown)


chldkdtn (approved):
It works for me after removing all the dependencies. @ckendrick can you add the instruction of compiling libROM in Lassen in README.md. Also please add your name under "Authors." Actually can you also add William Anderson and Paul Tranquilli as well?

chldkdtn (approved):

dylan-copeland (approved):
Thanks @ckendrick, this worked for me both on Lassen and another machine with different compilers. My only suggestion is that the example build with the flag `-d` is for a debug build. Either `-d` should be removed, or it should be commented that this is an example for building in debug mode.

ckendrick:
@dylan-copeland thanks for catching that. I just updated the wiki page to remove the debug option.

PR #261, opened 2024-01-17
Removal of time interval and closing files within BasisWriter::writeBasis

dreamer2368:
## Closing files at the end of `BasisWriter::writeBasis`

This PR revamps #150 . Now closing the database file is not associated with the destruction of `BasisWriter`. This enables writing both the basis and the snapshot at a given moment. As far as I believe, this should keep the backward compatibility as well, but it needs other people's review as we do not have a regression test to check this.

- `Database` class prints out the message when creating/opening a file.
- `BasisWriter` class initializes/destroys the member `Database* d_database` and `Database* d_snap_database` at its initialization/destruction, respectively. This, however, does not involve a file creation/opening/close.
- The `putInteger` operation in `~BasisWriter` is moved into `BasisWriter::writeBasis`.
- `BasisWriter::writeBasis` now closes the file at the end of the function call.

## Removal of time interval

The concept of time interval is obsolete, and it hasn't been used in practice. This is removed, addressing #267 .
While the inside of all functions do not use the concept of time intervals, some function signatures remain in order to keep the backward compatibility:
- `BasisGenerator::takeSample` used `double time` only for determining the time interval. Now `time` is a dummy variable and not used in the function.
  - In our practice, incremental svd's next sample is determined outside of this function.
- Member functions of `BasisReader` used `double time` for determining the time interval, but now `time` is a dummy variable.
  - `getSpatialBasis`
  - `getTemporalBasis`
  - `getSingularValues`
  - `getSnapshotMatrix`

~~The concept of time interval is obsolete, and it hasn't been used in practice. This will be removed in future (#267 ), and for now we enforce the number of interval not to be more than 1.~~

- ~~`Options` class will return error if `max_time_intervals > 1`. This will be kept until we actually remove the time interval concept.~~
- ~~`BasisWriter::writeBasis` assumes to newly create the file every time it is called.~~
  - ~~`HDF5Database::putIntegerArray` does not overwrite the dataset. If the dataset `num_time_intervals` is attempted to be re-written again, then an error will be returned.~~


lib/linalg/BasisWriter.cpp
```diff
@@ -81,13 +80,11 @@ BasisWriter::writeBasis(const std::string& kind)
     sprintf(tmp, "time_%06d", d_num_intervals_written);
```
dylan-copeland, line 80: In this class, `d_num_intervals_written` is output to the file in the `basis` case, not the `snapshot` case, and it simply counts the number of times a `basis` has been written. On the other hand, the class `BasisGenerator` has another concept of "intervals", which I assume should be related to the interval concept in `BasisWriter`. As far as I know, there is no use case of the interval concept in practice, although the code was written with the intent of some use case. I guess the idea was that after a certain number of intervals is reached in `BasisGenerator`, the `BasisWriter` should write the basis, and the `BasisGenerator` should go to the next interval. We don't use these classes in this way, and this design is not tested, so it may even be broken. In any case, it may be best to remove the interval concept from both classes.

dreamer2368: I've looked through the examples, and indeed there is no case of using more than 1 interval. Does this mean that time-windowing has been implemented differently, not using this `interval` concept?

dylan-copeland: Yes, we do have a time-windowing method that is implemented in applications and examples (e.g. Laghos), but it is not part of the libROM library. Time-windowing always uses 1 "interval" per time window. The determination to end a window is made in the application, not in libROM, since it should be problem-dependent. We do not use a general way to define windows that would be suitable for the library in general. When a window is ended, we destroy the `BasisGenerator` and construct a new one for the next window, never using the interval concept.

dreamer2368:
@dylan-copeland ,
After looking throughout the code, I'd like to suggest that we work on a separate PR for removing the concept of time intervals.

Even though the multiple time intervals are never used in practice, `num_time_intervals` or `d_num_intervals_written` are widely used in many classes and dataset names in hdf5 format. Since this concept of time interval has been so engrained into the code, it will take quite some time to remove it while keeping the consistency of the code. We will also have to discuss how we change the format of dataset name, which will have quite a big impact in backward compatibility.

dreamer2368:
@dylan-copeland , per our discussion, I added a commit that spits an error message whenever a user attempts to increase time interval, directly or indirectly through `BasisGenerator` or `SVD` class. Technically it still returns an error without this message, but this message will guide the user to increase the maximum number of samples.

Now `unit_tests/test_SVD.cpp` is removed from ci workflow. This test checks nothing but the concept of time interval, which is obsolete from now on.

I thought about the data format of basis classes, and ended up keeping the current implementation, allowing both csv and hdf5 format. Currently both data formats are compatible, and no reason to remove a functionality that we can support without any complication.

With that, I think this PR is ready for review and merge.

lib/linalg/BasisWriter.cpp
```diff
@@ -53,20 +53,19 @@ BasisWriter::BasisWriter(
...
-        d_database->putInteger("num_time_intervals", d_num_intervals_written);
-        d_database->close();
-        delete d_database;
-    }
-    if (d_snap_database) {
-        d_snap_database->putInteger("num_time_intervals", d_num_intervals_written);
-        d_snap_database->close();
-        delete d_snap_database;
-    }
+    if (d_database) delete d_database;
```
dylan-copeland, line 67: There is no need to check `if (d_database)`, since it is initialized by the constructor to `NULL` (which could be changed to `nullptr`) and can be deleted in any case.

lib/linalg/BasisWriter.cpp
```diff
@@ -122,14 +116,11 @@ BasisWriter::writeBasis(const std::string& kind)
 
         ++d_num_intervals_written;
 
+        d_database->putInteger("num_time_intervals", d_num_intervals_written);
```
dylan-copeland, line 119: Verify exactly 1 interval written?

lib/linalg/Options.h
```diff
@@ -42,15 +44,23 @@ class Options
...
         max_basis_dimension(samples_per_time_interval_),
         samples_per_time_interval(samples_per_time_interval_),
         max_time_intervals(max_time_intervals_),
         update_right_SV(update_right_SV_),
-        write_snapshots(write_snapshots_) {};
+        write_snapshots(write_snapshots_)
+    {
+        if (max_time_intervals > 1)
+        {
+            printf("time interval is obsolete and will be removed in the future. Set max_time_intervals=%d to 1!\n",
```
dylan-copeland, line 59: This line can be wrapped to make it shorter.

lib/linalg/svd/SVD.h
```diff
@@ -166,8 +166,9 @@ class SVD
     {
         int num_time_intervals =
             static_cast<int>(d_time_interval_start_times.size());
-        CAROM_VERIFY(d_max_time_intervals == -1 ||
-                     num_time_intervals < d_max_time_intervals);
+        CAROM_VERIFY(d_max_time_intervals == 1);
+        if (num_time_intervals > 0)
+            CAROM_ERROR("SVD::increaseTimeInterval- Time interval is obsolete and will be removed in the future. You received this error presumably because the number of samples reached its limit.\n");
```
dylan-copeland, line 171: Wrap to shorten.

lib/utils/CSVDatabase.h
```diff
@@ -47,7 +47,7 @@ class CSVDatabase : public Database
     virtual
```
dylan-copeland, line 47: It is good to add `override`, and `virtual` can be removed from all the member functions in this derived class, as well as other classes. If you prefer to make a future PR for that, that is fine with me.

lib/utils/Database.cpp
```diff
@@ -22,4 +31,18 @@ Database::~Database()
 {
 }
 
+bool
+Database::create(const std::string& file_name)
+{
+    std::cout << "Creating file: " << file_name << std::endl;
```
dylan-copeland, line 37: For these `cout` calls, should we store the MPI rank in the class and do `cout` only for rank 0?

dreamer2368: This would make sense for mpi-I/O, where multiple processors read/write a single file. Current implementation is file-per-processor, and Database uses a separate file name per processor. And each filename is printed out in the current implemeation.

This convention could be changed, which I think we can discuss aside from this PR. I'm actually in need of MPI-I/O, and planning to write an database class for it. Maybe we can update our data formats/print-out messages in the next PR.

dreamer2368:
@dylan-copeland ,

In my work for PR #274 , I ended up removing the concept of time interval. I moved those changes to here since it fits more to this PR.

While the inside of all functions do not use the concept of time intervals, some function signatures remain in order to keep the backward compatibility:
- `BasisGenerator::takeSample` used `double time` only for determining the time interval. Now `time` is a dummy variable and not used in the function.
  - In our practice, incremental svd's next sample is determined outside of this function.
- Member functions of `BasisReader` used `double time` for determining the time interval, but now `time` is a dummy variable.
  - `getSpatialBasis`
  - `getTemporalBasis`
  - `getSingularValues`
  - `getSnapshotMatrix`

Not exactly sure if we'd want to change these function signatures right away, which will impact all applications using libROM.  Need more discussion about this @chldkdtn @ckendrick

lib/linalg/Options.h
```diff
@@ -41,16 +43,24 @@ class Options
      *
      */
     Options(int dim_,
-            int samples_per_time_interval_,
-            int max_time_intervals_ = -1,
+            int max_num_samples_,
+            int max_time_intervals_ = 1,
```
dylan-copeland, line 47: Can this input be removed?

lib/linalg/svd/SVD.h
```diff
@@ -151,34 +115,27 @@ class SVD
      *         interval.
      */
     bool
-    isNewTimeInterval() const
+    isNewSample() const
```
dylan-copeland, line 118: Do we need this function? I'm not sure it makes sense without time intervals. It seems to be used only internally to check whether internal data has been initialized, so maybe it can be a protected function with a name like "isFirstSample"?

lib/linalg/svd/StaticSVD.h
```diff
@@ -128,9 +125,9 @@ class StaticSVD : public SVD
      *         date.
      */
     bool
-    thisIntervalBasisCurrent()
+    isBasisCurrent()
```
dylan-copeland, line 128: The comments for this function are not "current." It would be good to explain what "current" means.

lib/linalg/svd/SVD.h
```diff
@@ -151,34 +115,27 @@ class SVD
...
-        CAROM_VERIFY(d_max_time_intervals == -1 ||
-                     num_time_intervals < d_max_time_intervals);
-        d_time_interval_start_times.resize(
-            static_cast<unsigned>(num_time_intervals) + 1);
+        return d_num_samples;
     }
 
     /**
-     * @brief Get the number of samples taken.
+     * @brief Get the maximum number of samples that can be taken.
```
dylan-copeland, line 133: Comment on what happens if that maximum number of samples is exceeded?

lib/linalg/svd/StaticSVD.cpp
```diff
@@ -100,46 +100,27 @@ void StaticSVD::delete_factorizer()
...
 
     // Check the u_in is not non-zero.
     Vector u_vec(u_in, d_dim, true);
     if (u_vec.norm() == 0.0) {
         return false;
     }
 
-    if (isNewTimeInterval()) {
+    if (isNewSample()) {
         // We have a new time interval.
```
dylan-copeland, line 117: Comments about time intervals should be removed.

lib/linalg/BasisWriter.cpp
```diff
@@ -53,96 +52,77 @@ BasisWriter::BasisWriter(
...
 BasisWriter::writeBasis(const std::string& kind)
 {
-
     CAROM_ASSERT(kind == "basis" || kind == "snapshot");
 
     char tmp[100];
-    double time_interval_start_time =
-        d_basis_generator->getBasisIntervalStartTime(d_num_intervals_written);
-    sprintf(tmp, "time_%06d", d_num_intervals_written);
+    // This 0 index is the remainder from time interval concept.
```
dylan-copeland, line 74: Why can't we remove the 0 index now that we decided not to preserve backward compatibility for time intervals?

dreamer2368: Backward compatibility can be kept/removed at various level. We decide to break the code compatibility, so all the softwares depending on libROM should update their codes in order to build/compile. But I intended to keep data compatibility, so that all the previous data can be used seamlessly without additional conversion process.

I guess we can come up with a python script for dataset name conversion. Would that be a better path to you?

dylan-copeland: I'm not sure it is common practice to store data long term, although in theory someone may be doing that. I suppose most users generate FOM data, build ROMs, and test, without saving data for use in the future. In the rare case that someone is saving a lot of data, your suggestion to rename with a script should suffice, and we can leave that to the users. I would prefer that to maintaining indices in the code that no longer make sense.

dreamer2368: > I'm not sure it is common practice to store data long term, although in theory someone may be doing that. 

At least I'm the one keeping the snapshot and basis data. But I'm okay with having a format converter, so open to either route.

dreamer2368:
Dataset names for basis are now updated, removing the trailing 6 digits of integer that used to indicate time interval.

Any out-dated snapshot/basis hdf5 files can be updated using `scripts/data/update_basis_formats.py`. The syntax for using this python script is:
```
python3 update_basis_format.py filename
```

This basis/snapshot format update is tested with `unit_tests/test_basis_conversion.cpp`. In `.github/workflows/run_tests/action.yml`, outdated basis/snapshot files are first converted with the python script, and checked if the `BasisReader` class can read the updated files.

.github/workflows/run_tests/action.yml
```diff
@@ -4,7 +4,6 @@ runs:
     - name: Run unit tests
       run: |
           cd ${GITHUB_WORKSPACE}/build
-          ./tests/test_SVD
```
dylan-copeland, line 7: Does this remove the running of all SVD tests?

lib/linalg/BasisGenerator.cpp
```diff
@@ -129,35 +127,31 @@ BasisGenerator::isNextSample(
...
-            }
-            else {
-                d_basis_writer->writeBasis("basis");
-            }
-        }
-    }
-
-    return d_svd->takeSample(u_in, time, add_without_increase);
+    /*
+        Note for previous implementation:
```
dylan-copeland, line 143: I think it is better not to document previous implementations. It is simpler just to remove `resetDt`.

lib/linalg/BasisReader.h
```diff
@@ -58,92 +58,54 @@ class BasisReader {
...
-     *
-     * @param[in] time Time at which we are interested in the basis vectors.
-     *
      * @return True if the basis vectors at the requested time are different
      *         from the last requested basis vectors.
      */
     bool
-    isNewBasis(
-        double time)
+    isNewBasis()
```
dylan-copeland, line 65: Should this function and `d_last_basis_idx` be removed?

lib/linalg/svd/SVD.h
```diff
@@ -111,38 +108,25 @@ class SVD
...
+    int getNumSamples() const
     {
-        return static_cast<int>(d_time_interval_start_times.size());
+        return d_num_samples;
     }
 
     /**
-     * @brief Returns the start time for the requested time interval.
+     * @brief Get the maximum number of samples that can be taken.
+     *        SVD class will return an error if the number of samples exceed the maximum.
```
dylan-copeland, line 121: exceeds

CMakeLists.txt
```diff
@@ -270,12 +270,12 @@ if(GTEST_FOUND)
     GNAT
     QDEIM
     S_OPT
-    SVD
```
dylan-copeland, line 273: Why is `SVD` removed?

.github/workflows/run_tests/action.yml
dreamer2368: No it is not. This only tests the base `SVD` class, and only the time-interval related functionalities. The entire test routines become unnecessary as we remove the time interval.

lib/linalg/BasisReader.h
dreamer2368: I know this function is not used any more within libROM, but it wouldn't hurt to leave this function. I think the names can be confusing, so we can rename them, maybe `isFileLoaded()` and `bool d_file_loaded` ?

CMakeLists.txt
dreamer2368: As mentioned above. `test_SVD` only tests the base SVD class, and only the time-interval related functionalities. The entire test routines become unnecessary as we remove the time interval.

lib/linalg/BasisReader.h
dreamer2368: This makes sense and is reflected to PR.

lib/linalg/svd/IncrementalSVD.h
```diff
@@ -118,8 +115,7 @@ class IncrementalSVD : public SVD
     virtual
```
ckendrick, line 115: Mentions of `time` should be deleted from the doxygen comment at line 110 and 113.

lib/utils/Database.cpp
```diff
@@ -22,4 +31,18 @@ Database::~Database()
...
+}
+
+bool
+Database::open(
```
ckendrick, line 41: Is there a particular reason why a concrete implementation for the `open()` and `create()` functions were added to the abstract Database class? 
If these are needed, then perhaps they should return `true` by default since I noticed a compiler warning about no return type:
```
/home/test/libROM/lib/utils/Database.cpp: In member function 'virtual bool CAROM::Database::create(const string&)':
/home/test/libROM/lib/utils/Database.cpp:38:1: warning: no return statement in function returning non-void [-Wreturn-type]
   38 | }
```
It might be good in the future to try and unify some code here such as checking if the file exists on `open()` / does not exist on `create()`, etc, if those logic checks are shared between HDF/CSV Database classes, but that is certainly work for another PR.

ckendrick (changes requested):

lib/utils/Database.cpp
ckendrick: Disregard my comment above. I just noticed those changes in your next PR #274.

dreamer2368: Yeah I didn't realize this was not implemented here but in PR #274. Thanks for catching that.

dreamer2368:
@ebchin ,

After some discussion throughout this PR, we decided to change the dataset names within snapshot/basis hdf5 files. I heard from @dylan-copeland that you might have data files for your production and your simulation can be impacted by this update. I prepared a python script `scripts/data/update_basis_formats.py`, which can update the dataset names of the out-dated hdf5 snapshot/basis files. This script is also tested in the ci workflow. Please feel free to bring up any suggestion/concerns.

ckendrick (approved):

PR #263, opened 2024-01-23
scalapack bug fix for factorize

dreamer2368:
The scalapack routine `pdgesvd` performs a workspace query at its initial call in `factorize` in `lib/linalg/scalapack_f_wrapper.f90`, where it returns the required size of working array. If the size exceeds `1e9`, then an overflow occurs and the returned value can be invalid. 

The fortran routine `correct_svd_workarray_size` copied the size calculation part of scalapack `pdgesvd`, and computes the size as double in order to avoid the overflow. This routine is called in `factorize`, where it returns a warning message and resize the working array if the size seems to have experienced an overflow.

### DIFF
--- a/lib/linalg/scalapack_f_wrapper.f90
+++ b/lib/linalg/scalapack_f_wrapper.f90
@@ -462,6 +462,86 @@ subroutine create_local_matrix(A, x, m, n, src) bind(C)
     endif
 end subroutine
 
+subroutine correct_svd_workarray_size(A, desca, descu, descv, dou, dov, work)
+    use mpi
+
+    type(SLPK_Matrix), pointer, intent(in) :: A
+    integer, intent(in) :: desca(9), descu(9), descv(9)
+    character, intent(in) :: dou, dov
+    real(REAL_KIND), allocatable, intent(inout) :: work(:)
+
+    integer, parameter :: CTXT_=2, MB_=5, NB_=6, RSRC_=7, CSRC_=8
+
+    integer :: mrank, ierr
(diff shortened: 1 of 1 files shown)


ckendrick (commented):
@dreamer2368 have you verified if the output of `pdgesvd` is correct with these changes? This fix should allocate the proper workspace size, but I worry that `lwork` passed into `pdgesvd` has still overflowed and any integers Scalapack uses internally for indexing will overflow at these sizes.

We may have to modify the libROM scalapack wrapper to support int64 and compile [scalapack with support for 8 byte integers](https://github.com/Reference-ScaLAPACK/scalapack/blob/master/README#L139-L145) since it seems to be using int32 by default.

dreamer2368:
> @dreamer2368 have you verified if the output of `pdgesvd` is correct with these changes? This fix should allocate the proper workspace size, but I worry that `lwork` passed into `pdgesvd` has still overflowed and any integers Scalapack uses internally for indexing will overflow at these sizes.

I think you're right. I checked the resulting basis through running a regression test and it seems to be invalid. 

> We may have to modify the libROM scalapack wrapper to support int64 and compile [scalapack with support for 8 byte integers](https://github.com/Reference-ScaLAPACK/scalapack/blob/master/README#L139-L145) since it seems to be using int32 by default.

I'm not so clear how to implement this part and it would take some time to figure out. Meanwhile, I'll put this PR as WIP.

dreamer2368:
This workaround actually does not circumvent the overflow issue of scalapack. While we could compile both librom and scalapack with long integer, an ultimate solution is to run `factorize` in parallel. closing the PR now.

PR #268, opened 2024-02-11
A fix for StaticSVD::ComputeSVD to preserve snapshots

dreamer2368:
`StaticSVD::ComputeSVD` performs SVD using scalapack's `pdgesvd`, which does not preserve the original snapshot matrix. This can be a potential source of error when running `BasisGenerator::getSnapshotMatrix` after SVD operation (`BasisGenerator::endSamples()`).

`StaticSVD` now has an option to preserve the snapshot. This option is driven by the `bool` option `Options::static_svd_preserve_snapshot`. 

If snapshot preserving option is false, `StaticSVD::computeSVD` uses the original `d_samples` as before, modifying the matrix. `StaticSVD::getSnapshotMatrix` after svd operation will return an error, so that users would not use the modified snapshot matrix.

If snapshot preserving option is true, `StaticSVD::computeSVD` copies from `d_samples` into `snapshot_matrix` either transposed or original, and use `snapshot_matrix` as working matrix for SVD operation, preserving the original snapshot matrix `d_samples`. The working matrix `snapshot_matrix` is destroyed at the end of the function call.

### DIFF
--- a/lib/linalg/Options.h
+++ b/lib/linalg/Options.h
@@ -345,6 +345,11 @@ public:
      *        incremental SVD algorithm.
      */
     double max_sampling_time_step_scale = 5.0;
+
+    /**
+     * @brief Option to preserve snapshot in StaticSVD::computeSVD.
+     */
+    bool static_svd_preserve_snapshot = false;
 };
 
 }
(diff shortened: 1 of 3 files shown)


dreamer2368:
@chldkdtn @dylan-copeland @ckendrick @siuwuncheung , I made the change per @dylan-copeland 's suggestion and it is ready for review and merge.

I made this PR separate from other branches so that this can be merged independently.

ckendrick (approved):

dylan-copeland (approved):

Issue #269, opened 2024-02-13
Hard-coded values in NNLS solver


dreamer2368:
Line 66 and 163 of `lib/linalg/NNLS.cpp` use hard-coded values of `n_proc_max_for_partial_matrix_ = 15` and `nb = 3`, which seem to be used mainly for distributing the matrix.

- There is no sufficient description as to why these values are chosen.
- For better scalability of this solver, shouldn't these values be determined by the users?

PR #271, opened 2024-02-15
minor fix and a unit test for NNLSSolver

dreamer2368:
### `n_dist_loc_max` fix
In `NNLSSolver::solve_parallel_with_scalapack`, `n_dist_loc_max` is now determined from `n_tot`, not `m` (as it is supposed to). `n_dist_loc_max` only determines the memory size of the working array, not impacting the NNLS solution. Before this fix, the following working array variables were allocated with unnecessarily large sizes:
```
// The following matrices are stored in column-major format as Vectors
Vector mat_0_data(m * n_dist_loc_max, false);
Vector mat_qr_data(m * n_dist_loc_max, false);
```

### hard-coded limit for number of processors
Per issue #269 , the number of processors that can be used for NNLSSolver is limited to `15`. This limit is removed and tested through the unit test below.

### termination criterion for NNLSSolver
Previously, NNLS iteration is terminated when $L_{\infty}$-norm (the maximum value) of the residual vector is below the threshold. Now NNLSSolver has an option to terminate when $L_2$-norm of the residual vector is below a corresponding threshold. This is set at initialization with the input argument `NNLS_termination criterion`.

## unit test for NNLSSolver

A unit test for `NNLSSolver::solve_parallel_with_scalapack` is added.
This checks the solution approximates the system with desired tolerance, both in serial and parallel.

On LC quartz, the scaling test result is: (for `300x1000`)
| Number of processors  | Time     |
| :---        |          ---: |
| 1      | 908 ms  |
| 2     | 769 ms   |
| 4     | 563 ms   |
| 8     | 469 ms   |
| 12     | 463 ms   |
| 15     | 456 ms   |

Another test with a larger system (`500x15000`):
| Number of processors  | Time     |
| :---        |          ---: |
| 1      | 24216 ms  |
| 2     | 13287 ms   |
| 4     | 7843 ms   |
| 8     | 5091 ms   |
| 12     | 4162 ms   |
| 15     | 3666 ms   |

This timing includes the entire setup for matrix-vector system aside from `solve_parallel_with_scalapack`.


lib/linalg/NNLS.h
```diff
@@ -19,6 +19,16 @@
 
 namespace CAROM {
 
+/**
+ * @brief Termination criterion for NNLS solver:
+ *        LINF: L-infinity norm (maximum value) of the residual
+ *        L2: L2 norm of the residual
+ */
+enum NNLS_termination {
```
dylan-copeland, line 27: `enum` -> `enum class`

unit_tests/test_NNLS.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+TEST(NNLS, solve_with_LINF)
+{
+    int nproc;
+    int rank;
+    MPI_Comm_size(MPI_COMM_WORLD, &nproc);
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+
+    const int nrow = 300;
+    const int ncol = 1000;
```
dylan-copeland, line 38: This is a rather large test size. Is the test slow? If so, would it be sufficient to test with a smaller system?

unit_tests/test_NNLS.cpp
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+
+    std::cout << rank << ": Number of nonzeros in NNLS solution: " << nnz
+              << ", out of " << rom_sol.dim() << std::endl;
+
+    MPI_Allreduce(MPI_IN_PLACE, &nnz, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
+
+    if (rank == 0)
+        std::cout << "Global number of nonzeros in NNLS solution: " << nnz << std::endl;
```
dylan-copeland, line 101: Should there be a check that `nnz` is less than some number based on the result of this test?

dreamer2368: Do you mean a prescribed answer based on the given random seed? Yeha I think it's possible.

dylan-copeland: Yes, I think the test would be stronger if it also checks `nnz`, and it would be necessary to fix the random seed to make the test deterministic.

dylan-copeland (approved):

lib/linalg/NNLS.cpp
```diff
@@ -58,14 +58,15 @@ extern "C" {
 NNLSSolver::NNLSSolver(double const_tol, int min_nnz, int max_nnz,
                        int verbosity,
                        double res_change_termination_tol,
-                       double zero_tol, int n_outer, int n_inner)
+                       double zero_tol, int n_outer, int n_inner,
+                       const NNLS_termination criterion)
```
ckendrick, line 62: It might be good to have a `CAROM_VERIFY` check in the constructor to make sure that `NNLS_termination` is set to `LINF` or `L2`.

ckendrick (approved):

PR #274, opened 2024-02-17
MPIO support for Basis classes

dreamer2368:
This PR intends to support multi-path input/output (MPIO) of HDF5 format, thereby enabling MPI I/O of a single file. 

Currently the I/O only supports 'file-per-processor' parallel I/O, where each processor reads/writes its own file. While this guarantees the I/O scalability most of the time, this fixates the number of processors for the entire workflow of basis generation and continual use.

Single file I/O with multiple processors, of course, can cause a significant overhead. This, however, has been optimized within HDF5 library up to some extent. As long as the file system is appropriately set up (e.g. Lustre stripe count), the parallel I/O on a single file can maintain reasonable time cost (from [HDF5 group post](https://www.hdfgroup.org/2015/08/parallel-io-with-hdf5/)).
If used carefully, this option will provide users with the freedom of changing number of processors throughout their own basis generation/use workflow.

List of implementation:
- [x] `HDFDatabaseMPIO` class
- [x] Deploying `HDFDatabaseMPIO` as an option in `BasisGenerator`, `BasisReader`, and `BasisWriter`
- [x] ~~(optional) removing the concept of time interval in basis data formats.~~

Comparison of scalability for `5000x1000` snapshot matrix, between `HDFDatabase` and `HDFDatabaseMPIO`:
- `BasisGenerator::endSamples` (SVD + basis writing)

| Number of processors  | Base     |  MPIO    |
| :---        |  :---:  |         ---: |
| 1      | 10352ms  |  10253ms  |
| 2     | 2131ms   |  2150ms   |
| 4     | 1548ms   |  1605ms  |
| 8     | 1566ms   |  1524ms  |

- `BasisGenerator::writeSnapshots` (snapshot writing only)

| Number of processors  | Base     |  MPIO    |
| :---        |  :---:  |         ---: |
| 1      | 232ms  |  246ms  |
| 2     | 213ms   | 200ms  |
| 4     | 171ms   |  175ms  |
| 8     | 198ms   |  178ms |

The comparison is done on LC quartz debug node, with lustre stripe count of 36.


lib/linalg/BasisReader.h
```diff
@@ -197,6 +200,11 @@ class BasisReader {
      *
      * @brief Returns the dimension of the system on this processor.
      *
+     * @param[in] kind  The target matrix of which dimension (row) is returned.
```
dylan-copeland, line 203: This comment is not clear to me. Would this be better? `Type of matrix whose row-dimension is returned.`

lib/linalg/BasisReader.cpp
```diff
@@ -135,7 +151,7 @@ BasisReader::getTemporalBasis()
     int num_cols = getNumSamples("temporal_basis");
 
     char tmp[100];
-    Matrix* temporal_basis_vectors = new Matrix(num_rows, num_cols, true);
+    Matrix* temporal_basis_vectors = new Matrix(num_rows, num_cols, false);
```
dylan-copeland, line 154: Why is this changed to non-distributed? Was it a bug for it to be distributed in the past?

lib/utils/CSVDatabase.h
```diff
@@ -84,7 +85,6 @@ class CSVDatabase : public Database
      * @param[in] data The array of integer values to be written.
      * @param[in] nelements The number of integers in the array.
      */
-    virtual
```
dylan-copeland, line 87: It looks like a lot of functions like this need `override`?

lib/utils/Database.h
```diff
@@ -258,8 +260,11 @@ class Database
      * @param[out] data The allocated array of double values to be read.
      * @param[in] nelements The number of doubles in the array.
      * @param[in] offset The initial offset in the array.
+     *                   Typically, this is a column index of the matrix data.
```
dylan-copeland, line 263: Would it be clearer to say `zero-based column index`?

lib/utils/Database.h
```diff
@@ -277,9 +282,134 @@ class Database
      */
     enum formats {
         HDF5,
-        CSV
+        CSV,
```
dylan-copeland, line 285: `enum class`

lib/utils/Database.h
```diff
@@ -277,9 +282,134 @@ class Database
...
+     *        For HDFDatabaseMPIO, the global array is written into a single file via MPI I/O.
+     *
+     * @param[in] key The key associated with the array of values to be
+     *                written.
+     * @param[in] data The array of double values to be written.
+     * @param[in] nelements The number of doubles in the array.
+     */
+    virtual
+    void
+    putDoubleArray_parallel(
```
dylan-copeland, line 350: Would the interface be simpler if we just use the function name `putDoubleArray`, with an optional input for the type?

lib/utils/HDFDatabase.h
```diff
@@ -227,19 +231,129 @@ class HDFDatabase : public Database
...
+     *
+     * @param[in] file_name Name of HDF5 database file to create.
+     * @param[in] comm      MPI communicator to obtain the process rank.
+     *
+     * @return True if file create was successful.
      */
-    HDFDatabase(
-        const HDFDatabase& other);
+    virtual bool
+    create_parallel(
```
dylan-copeland, line 244: Would the interface be simpler if we just add optional inputs to the functions `create`, `open`, `close`, rather than adding new function names ending in `_parallel`? If so, maybe the `_parallel` functions could be kept as private functions and called by the functions without `_parallel` in the parallel case.

lib/utils/HDFDatabase.h
```diff
@@ -313,6 +428,26 @@ class HDFDatabase : public Database
      * @brief The key representing an integer array.
      * */
     static const int KEY_INT_ARRAY;
+
+    /**
+     * @brief MPI process rank.
+     *        Used only when d_parallel is true.
```
dylan-copeland, line 434: Where is `d_parallel`?

lib/utils/HDFDatabaseMPIO.cpp
```diff
@@ -0,0 +1,627 @@
+/******************************************************************************
+ *
+ * Copyright (c) 2013-2023, Lawrence Livermore National Security, LLC
```
dylan-copeland, line 3: 2024

unit_tests/test_HDFDatabase.cpp
```diff
@@ -0,0 +1,867 @@
...
+    printf("rank: %d, duration: %dms\n", rank, duration);
+
+    for (int d = 0; d < nrow_local * ncol; d++)
+        EXPECT_TRUE(data[d] == d + offsets[rank] * ncol);
+}
+
+TEST(HDF5, Test_selective_parallel_writing)
+{
+    int nproc, rank;
+    MPI_Comm_size(MPI_COMM_WORLD, &nproc);
```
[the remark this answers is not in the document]
dreamer2368: This test should work for general number of processors, and we just use 3 processors in ci workflow without a specific reason.

lib/linalg/BasisReader.cpp
dreamer2368: I do think it was a bug in the past. When I'm looking at `ComputeSVD`, temporal basis matrix is clearly stored as a non-distributed matrix (each process owns the global matrix)

lib/utils/Database.h
dreamer2368: I guess we could add an optional input `bool distributed=false`. This change might impact the existing applications again, but I agree this would be simpler in the long run.

lib/utils/HDFDatabase.h
dreamer2368: We'd probably have to put the same optional input for `put...Array` and `get...Array` functions. I agree this would be a simpler interface to users in general.

lib/utils/HDFDatabase.h
dreamer2368: This description should be corrected.

unit_tests/test_HDFDatabase.cpp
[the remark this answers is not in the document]
dylan-copeland: Then should we test for `nproc >= 3`? It looks like it would fail otherwise.

lib/utils/CSVDatabase.h
dreamer2368: `Database::putIntegerArray = 0` is a pure virtual function. Do we still need explicit `override`?

dylan-copeland: The `override` may not be necessary, but my understanding is that it is used for clarity and to prevent bugs. By adding `override` to a function, the compiler will check that the function matches a virtual function. This prevents bugs like typos in the function name, different argument types (including references, pointers), etc. So it is more a matter of style and good practice than necessity.

lib/linalg/BasisReader.h
```diff
@@ -44,10 +44,13 @@ class BasisReader {
      * @param[in] db_format Format of the file to read.
      *                      One of the implemented file formats defined in
      *                      Database.
+     * @param[in] dim Dimension of the basis that will be read from a file.
```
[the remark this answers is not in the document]
dreamer2368: For now, I only clarified `BasisReader.h`.

dreamer2368:
Reflecting @dylan-copeland 's comments, function signatures of database are changed:

- `open`, `create` now takes an optional input `MPI_Comm comm=MPI_COMM_NULL`, If `comm==MPI_COMM_NULL`, the file is opened/created serially.
- `put...Array` and `get...Array` functions take an optional input `bool distributed=false`. If `distributed`, different types of parallel I/O are performed depending on classes:
  - `CSVDatabase`: all I/O are performed serially whether the data is distributed or not.
  - `HDFDatabase`: Parallel I/O is performed as file-per-process, where each process opens/creates a file corresponding to its own rank.
  - `HDFDatabase`: all processes perform parallel I/O on a single file. If not `distributed`, then only the root process performs I/O.

lib/utils/Database.h
```diff
@@ -269,17 +302,143 @@ class Database
...
+    //  *        For HDFDatabase, the function is equivalent to create,
+    //  *        only extending the file name with 6 digits indicating processor rank.
+    //  *        For HDFDatabaseMPIO, the file is created with MPI I/O file access property.
+    //  *
+    //  * @param[in] file_name Base Name of CSV database file to create.
+    //  * @param[in] comm      MPI communicator to get the rank.
+    //  *
+    //  * @return True if file create was successful.
+    //  */
+    // virtual
```
dylan-copeland, line 330: Remove commented code?

lib/utils/HDFDatabase.h
```diff
@@ -225,21 +263,132 @@ class HDFDatabase : public Database
...
+    //  * @brief Creates a new HDF5 database file with the supplied name,
+    //  *        extending it with 6 digits indicating the process rank.
+    //  *
+    //  * @param[in] file_name Name of HDF5 database file to create.
+    //  * @param[in] comm      MPI communicator to obtain the process rank.
+    //  *
+    //  * @return True if file create was successful.
+    //  */
+    // virtual bool
+    // create_parallel(
```
dylan-copeland, line 279: Commented code.

lib/utils/HDFDatabaseMPIO.h
```diff
@@ -0,0 +1,455 @@
...
+    virtual
+    ~HDFDatabaseMPIO() {}
+
+    /**
+     * @brief Creates a new HDF5 database file for distributed data
+     *        with the supplied name.
+     *
+     * @param[in] file_name Name of HDF5 database file to create.
+     * @param[in] comm MPI communicator for distributed data I/O.
+     *                 HDFDatabaseMPIO does not allow MPI_COMM_NULL.
```
dylan-copeland, line 43: Comment about what the default communicator does?

lib/utils/HDFDatabaseMPIO.h
```diff
@@ -0,0 +1,455 @@
...
+    //  * @brief Creates a new HDF5 database file for non-distributed data
+    //  *        with the supplied name.
+    //  *        For HDFDatabaseMPIO, the behavior is equilvalent to create_parallel.
+    //  *
+    //  * @param[in] file_name Name of HDF5 database file to create.
+    //  *
+    //  * @return True if file create was successful.
+    //  */
+    // bool
+    // create(
```
dylan-copeland, line 62: Commented code.

lib/utils/HDFDatabase.h
```diff
@@ -225,21 +263,132 @@ class HDFDatabase : public Database
...
+    //  * @brief Opens an existing HDF5 database file with the supplied name,
+    //  *        extending it with 6 digits indicating the process rank.
+    //  *
+    //  * @param[in] file_name Name of existing HDF5 database file to open.
+    //  * @param[in] type      Read/write type ("r"/"wr")
+    //  * @param[in] comm      MPI communicator to obtain the process rank.
+    //  *
+    //  * @return True if file open was successful.
+    //  */
+    // virtual
```
ckendrick, line 304: There are a few places with some commented out blocks of code. Are those needed for something in the future, or can they be removed?

lib/utils/HDFDatabaseMPIO.h
```diff
@@ -0,0 +1,455 @@
...
+#define included_HDFDatabaseMPI_h
+
+#include "HDFDatabase.h"
+
+namespace CAROM {
+
+/**
+ * HDFDatabaseMPIO implements Multi-Path Input/Output (MPIO) of HDF5.
+ */
+class HDFDatabaseMPIO : public HDFDatabase
```
ckendrick, line 23: I noticed compiling this branch with the serial version of HDF5 fails. I saw you updated the Dockerfile dependencies to install the parallel version which is good, but should we support compiling with the serial version of HDF5 still? If so, we can modify this a bit to conditionally include the MPIO related things depending on whether the H5 library has parallel support using `H5_HAVE_PARALLEL`. It might require several ifdef statements around the code and in BasisReader/BasisWriter though.
If we don't want to do that now, then maybe we can make sure that parallel HDF5 is found by CMake. I think adding something like this to the main `CMakeLists.txt` after the find package statement for HDF5 should work:
```
if(NOT ${HDF5_IS_PARALLEL})
  message(FATAL_ERROR "HDF5 library does not support MPI - please use parallel HDF5")
endif()
```

dreamer2368:
Reflecting @ckendrick 's comment, parallel hdf5 is set optional. `HDFDatabaseMPIO` has the full capability only when parallel hdf5 is available. Otherwise, it becomes identical to its base clasee `HDFDatabase`.

dylan-copeland (approved):

ckendrick (approved):

PR #276, opened 2024-03-11
Fix issue with Database class

ckendrick:
This adds a return statement to the base Database functions `open()` and `create()`. Without these, calls to `endSamples()` and `writeSnapshot()` would hang indefinitely printing newlines to screen. 

This also adds a test to the CI that calls `writeSnapshot()` to catch this behavior in the future.

### DIFF
--- a/lib/utils/Database.cpp
+++ b/lib/utils/Database.cpp
@@ -16,12 +16,6 @@
 
 namespace CAROM {
 
-bool fileExists(const std::string& name)
-{
-    std::ifstream f(name.c_str());
-    return f.good();
-    // ifstream f will be closed upon the end of the function.
-}
 
 Database::Database()
 {
--- a/unit_tests/smoke_static.cpp
+++ b/unit_tests/smoke_static.cpp
@@ -144,5 +144,5 @@ main(
 
     // Finalize MPI and return.
     MPI_Finalize();
-    return !status;
+    return 0;
 }
--- a/unit_tests/smoke_static.cpp
+++ b/unit_tests/smoke_static.cpp
@@ -53,8 +53,6 @@ main(
         return 1;
     }
 
-    bool status = false;
-
     // Define the values for the first sample.
     double vals0[6] = {1.0, 6.0, 3.0, 8.0, 17.0, 9.0};
 
--- a/.github/workflows/run_tests/action.yml
+++ b/.github/workflows/run_tests/action.yml
@@ -7,6 +7,7 @@ runs:
           ./tests/test_Vector
           ./tests/test_Matrix
           mpirun -n 3 --oversubscribe ./tests/test_Matrix
+          ./tests/smoke_static
           ./tests/test_DEIM
           ./tests/test_GNAT
           ./tests/test_QDEIM
--- a/lib/linalg/BasisReader.cpp
+++ b/lib/linalg/BasisReader.cpp
@@ -103,7 +103,7 @@ BasisReader::getSpatialBasis(
     int num_cols = getNumSamples("basis");
 
     char tmp[100];
-    CAROM_VERIFY(0 < start_col <= num_cols);
+    CAROM_VERIFY(0 < start_col && start_col <= num_cols);
     CAROM_VERIFY(start_col <= end_col && end_col <= num_cols);
     int num_cols_to_read = end_col - start_col + 1;
 
@@ -177,7 +177,7 @@ BasisReader::getTemporalBasis(
     int num_cols = getNumSamples("temporal_basis");
 
     char tmp[100];
-    CAROM_VERIFY(0 < start_col <= num_cols);
+    CAROM_VERIFY(0 < start_col && start_col <= num_cols);
     CAROM_VERIFY(start_col <= end_col && end_col <= num_cols);
     int num_cols_to_read = end_col - start_col + 1;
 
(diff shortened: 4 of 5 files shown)


unit_tests/smoke_static.cpp
```diff
@@ -144,5 +144,5 @@ main(
 
     // Finalize MPI and return.
     MPI_Finalize();
-    return !status;
```
dylan-copeland, line 147: `status` is now an unused variable.

dylan-copeland:
@ckendrick can you please provide an example test that demonstrates the issue?

ckendrick:
@dylan-copeland this issue should occur if you run any example from the current upstream branch during the offline phase. This should happen to any code that calls `endSamples()` or `writeSnapshot()`. Some examples I tried were
`dg_advection_global_rom -offline`
`poisson_global_rom -offline -f 1.0 -id 0`

lib/utils/Database.cpp
```diff
@@ -35,6 +35,7 @@ bool
 Database::create(const std::string& file_name)
 {
     std::cout << "Creating file: " << file_name << std::endl;
+    return true;
```
dylan-copeland, line 38: The base class doesn't do anything. Wouldn't it make more sense to make this a pure virtual function?

dylan-copeland: Also, it looks like `fileExists` is unused and can be removed.

dreamer2368, line 38: I think this change is already made at #274 . I'm open to @dylan-copeland 's suggestion.

chldkdtn:
@dylan-copeland would you review this PR again?

dylan-copeland (approved):

chldkdtn:
@dreamer2368 do you approve this PR?

lib/utils/Database.cpp
```diff
@@ -31,23 +24,4 @@ Database::~Database()
 {
 }
 
-bool
-Database::create(
-    const std::string& file_name,
-    const MPI_Comm comm)
-{
-    std::cout << "Creating file: " << file_name << std::endl;
```
dreamer2368, line 39: This does not affect the I/O behavior, but we have been printing out a statement whenever creating/opening a file. Removing this function is fine, but we probably want to keep this printing statement where it is used.

lib/utils/Database.cpp
```diff
@@ -31,23 +24,4 @@ Database::~Database()
...
-    return true;
-}
-
-bool
-Database::open(
-    const std::string& file_name,
-    const std::string& type,
-    const MPI_Comm comm)
-{
-    std::cout << "Opening file: " << file_name << std::endl;
```
dreamer2368, line 49: Same as above

lib/utils/HDFDatabase.cpp
```diff
@@ -57,8 +57,6 @@ HDFDatabase::create(
         ext_filename += tmp;
     }
 
-    Database::create(ext_filename, comm);
```
dreamer2368, line 60: We would want to keep the printing statement for filename here.

lib/utils/HDFDatabase.cpp
```diff
@@ -89,8 +87,6 @@ HDFDatabase::open(
         ext_filename += tmp;
     }
 
-    Database::open(ext_filename, type, comm);
```
dreamer2368, line 92: Same as above

lib/utils/HDFDatabaseMPIO.cpp
```diff
@@ -45,8 +45,6 @@ HDFDatabaseMPIO::create(
 
     /* Create a single file with a name compatible to HDFDatabase. */
     std::string file_name_ext(file_name + ".000000");
-    if (d_rank == 0)
-        Database::create(file_name_ext);
```
dreamer2368, line 49: Same above

lib/utils/HDFDatabaseMPIO.cpp
```diff
@@ -97,8 +95,6 @@ HDFDatabaseMPIO::open(
         d_rank = 0;
 
     std::string file_name_ext(file_name + ".000000");
-    if (d_rank == 0)
-        Database::open(file_name_ext, type);
```
dreamer2368, line 101: Same above

lib/utils/CSVDatabase.cpp
```diff
@@ -31,7 +31,6 @@ CSVDatabase::create(
     const std::string& file_name,
     const MPI_Comm comm)
 {
-    Database::create(file_name, comm);
```
dreamer2368, line 34: Csv doesn't really create or open a file with this function, so I'm okay with removing the printing statement here

dreamer2368 (changes requested):

dreamer2368 (approved):

chldkdtn:
@dylan-copeland we need your review again on this PR.

unit_tests/smoke_static.cpp
```diff
@@ -144,5 +142,5 @@ main(
...
+    return 0;
```
dylan-copeland, line 145: It looks like the goal of this test just to run without failing? There is nothing checked, like in `test_Matrix.cpp` where `EXPECT*` functions are called. Is this the intention?

lib/utils/Database.cpp
```diff
@@ -16,12 +16,6 @@
...
 namespace CAROM {
 
-bool fileExists(const std::string& name)
-{
-    std::ifstream f(name.c_str());
-    return f.good();
-    // ifstream f will be closed upon the end of the function.
-}
 
 Database::Database()
```
dylan-copeland, line 20: This can go in the header, as well as the destructor, for easier reading.

dylan-copeland (approved):

unit_tests/smoke_static.cpp
ckendrick: Yes this test was added to the CI only intending to be a quick check if an error occurred with `endSamples()`/`writeSnapshot()`. The issue that caused this problem has since been fixed, but the only test I could find that uses these functions is in `test_HDFDatabase.cpp` which is restricted to parallel HDF5 builds only.
In the future, we should provide a serial test in `test_HDFDatabase.cpp` and remove this test (and other outdated/unused tests).

PR #288, opened 2024-05-18
Get R factor from QR factorization

dylan-copeland:


cval26 (approved):

ckendrick (changes requested):
This looks good but `test_QR` should probably get added to the CI.

ckendrick (approved):

PR #289, opened 2024-05-22
Snapshot dmd

ptranq:
Adds the SnapshotDMD class derived from DMD.  This class essentially performs DMD but interpolates the current snapshots to new "artificial snapshots" in large enough quantity to produce a DMD ROM with the given reduced dimension.  This interpolation is performed with the SnapshotInterpolator class, specifically with PCHIP interpolation.

The wave_equation example is extended to provide the option of using snapshotDMD, as well minor improvements to provide the option to use initial condition projection for windows after the first, and the option to provide an output directory to minimize clutter after running the example.


lib/algo/SnapshotDMD.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+    PCHIPInterpolator* interp = new PCHIPInterpolator();
+    std::vector<Vector*> new_snapshots;
+    std::vector<Vector*> new_times;
+
+    interp->interpolate(d_sampled_times,d_snapshots,n,new_times,new_snapshots);
+    d_snapshots = new_snapshots;
+    d_sampled_times = new_times;
+    d_dt = d_sampled_times[2]->getData()[0]-d_sampled_times[1]->getData()[0];
+
```
siuwuncheung, line 61: Redundant line

lib/algo/SnapshotDMD.h
```diff
@@ -0,0 +1,132 @@
...
+     * @param[in] state_offset d_state_offset
+     * @param[in] derivative_offset d_derivative_offset
+     */
+    SnapshotDMD(std::vector<std::complex<double>> eigs, Matrix* phi_real,
+                Matrix* phi_imaginary, int k,
+                double dt, double t_offset,
+                Vector* state_offset) :
+        DMD(eigs, phi_real, phi_imaginary, k, dt, t_offset, state_offset) {}
+
+
```
siuwuncheung, line 128: Redundant line

examples/dmd/wave_equation.cpp
```diff
@@ -558,6 +604,12 @@ int main(int argc, char *argv[])
 
             if (i % windowNumSamples == 0 && i < ts.size()-1)
             {
+                if(project_initial_condition)
+                {
+                    result_u = dmd_u[curr_window]->predict(ts[i]);
+                    cout << "Projecting solution for new window at " << ts[i] << endl;
+                    dmd_u[curr_window+1]->projectInitialCondition(result_u, ts[i]);
```
siuwuncheung, line 611: For me to better understand the changes, I guess this projection is not necessarily relevant to the snapshot interpolation?

examples/dmd/wave_equation.cpp
```diff
@@ -24,6 +24,19 @@
...
 // Elapsed time for predicting DMD: 2.496171e-03 second
+//
+// Sample run using snapshotDMD and using projected initial conditions for the
+//  predicition phase.
+//
+// Command 2:
+//  wave_equation -tf 2 -nwinsamp 25 -rdim 27 -visit -snap -proj
+//
+// Output 2:
+//  Relative error of DMD solution (u) at t_final: 2 is 0.0029577152
```
siuwuncheung, line 35: I realize that the error is higher than the original outputs, with a mixed effect of enabling snapshot interpolation and the projection of initial condition. It could be helpful if we know which one is the source of the increase of the error.

siuwuncheung (commented):
Looks good to me! An exciting enhancement to DMD. Thanks, @ptranq

lib/algo/SnapshotDMD.cpp
ptranq: Removed

lib/algo/SnapshotDMD.h
ptranq: Removed

examples/dmd/wave_equation.cpp
ptranq: Correct.  This is simply projecting the initial conditions in each subsequent window to come from the final conditions of the preceding window.  This makes the DMD solution be a prediction across the entire time domain, instead of a series of approximations each across only a single window.

examples/dmd/wave_equation.cpp
ptranq: I have updated the output for command 1 to reflect what I obtain on quartz.  I get a different error than what was reported previously while running on this branch, and confirmed the same (new) numbers on the master branch.  Additionally, I have added two new examples: the first performs the current "command 1" with projection enabled, and the second performs the current "command 2" with projection disabled.  The errors are much closer together and should, hopefully, clear things up.

examples/dmd/wave_equation.cpp
```diff
@@ -256,6 +297,15 @@ int main(int argc, char *argv[])
                    "Reduced dimension for DMD.");
     args.AddOption(&windowNumSamples, "-nwinsamp",    "--numwindowsamples",
                    "Number of samples in DMD windows.");
+    args.AddOption(&snapshotDMD, "-snap", "--snapshot-dmd", "-no-snap",
+                   "--no-snapshot-dmd",
+                   "Use snapshot DMD instead of standard DMD.");
+    args.AddOption(&project_initial_condition, "-proj", "--project-initial",
+                   "-no-proj",
+                   "--no-project-initial",
+                   "Project the initial condition in subsequent window as the final value of the current window.");
```
dylan-copeland, line 306: These long lines can be broken by using multiple `""`.

lib/algo/SnapshotDMD.h
```diff
@@ -0,0 +1,131 @@
...
+namespace CAROM {
+
+class Vector;
+class SnapshotDMD : public DMD
+{
+public:
+
+    /**
+     * @brief Constructor. Basic DMD with uniform time step size.
```
dylan-copeland, line 15: Should the description include something specific to `SnapshotDMD`? It looks like this comment is copied from another class.

lib/algo/SnapshotDMD.h
```diff
@@ -0,0 +1,131 @@
...
+     * @brief Returns a copy of the current snapshot vector "d_snapshots"
+     */
+    std::vector<Vector*> getSnapshotVectors()
+    {
+        std::vector<Vector*> return_snapshots(d_snapshots);
+        return return_snapshots;
+    }
+protected:
+    /**
+     * @brief Obtain DMD model interpolant at desired parameter point by
```
dylan-copeland, line 82: Since this is just declaring the existing function as a `friend`, I don't think we need to copy the comments here. They are already in ParametricDMD.h. Maybe just a 1-line comment like `See ParametricDMD.h for documentation.` The comments could be removed also from existing files like DMD.h and NonuniformDMD.h, since they are just duplicated. I wonder if there is a better design, perhaps with `getParametricDMD` being a method of the base `DMD` class, but we can leave that question for another PR.

examples/dmd/wave_equation.cpp
```diff
@@ -448,9 +501,17 @@ int main(int argc, char *argv[])
...
-    dmd_u[curr_window]->takeSample(u.GetData(), t);
+    if(snapshotDMD)
+    {
+        dmd_u.push_back(new CAROM::SnapshotDMD(u.Size(), dt));
+    }
+    else
+    {
+        dmd_u.push_back(new CAROM::DMD(u.Size(), dt));
+    }
+    dmd_u[curr_window]->takeSample(u, t);
```
dylan-copeland, line 514: `u.GetData()` (to avoid deprecated warning)

examples/dmd/wave_equation.cpp
```diff
@@ -487,8 +548,15 @@ int main(int argc, char *argv[])
...
-                dmd_u[curr_window]->takeSample(u.GetData(), t);
+                if(snapshotDMD)
+                {
+                    dmd_u.push_back(new CAROM::SnapshotDMD(u.Size(), dt));
+                }
+                else
+                {
+                    dmd_u.push_back(new CAROM::DMD(u.Size(), dt));
+                }
+                dmd_u[curr_window]->takeSample(u, t);
```
dylan-copeland, line 559: `u.GetData()` (to avoid deprecated warning)

examples/dmd/wave_equation.cpp
```diff
@@ -20,10 +20,45 @@
...
-// Elapsed time for predicting DMD: 2.496171e-03 second
+//  Relative error of DMD solution (u) at t_final: 5 is 0.00015906082
+//  Elapsed time for solving FOM: 2.223148e+00 second
+//  Elapsed time for training DMD: 1.140992e+00 second
+//  Elapsed time for predicting DMD: 1.082472e-02 second
+//
+// Sample run with Time-Windowing DMD with initial condition projection
+//
+// Command 2:
+//  wave_equation -0 4 -tf 5 -nwinsamp 25 -proj
```
dylan-copeland, line 31: `-0`

examples/dmd/wave_equation.cpp
```diff
@@ -20,10 +20,45 @@
...
+//
+// Output 2:
+//  Relative error of DMD solution (u) at t_final: 5 is 0.029780998
+//  Elapsed time for solving FOM: 2.348481e+00 second
+//  Elapsed time for training DMD: 1.141874e+00 second
+//  Elapsed time for predicting DMD: 1.075082e-02 second
+//
+// Sample run using snapshotDMD and using projected initial conditions for the
+//  predicition phase.
```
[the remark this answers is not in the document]
ptranq: Fixed

examples/dmd/wave_equation.cpp
```diff
@@ -20,10 +20,45 @@
...
+//  wave_equation -tf 2 -nwinsamp 25 -rdim 27 -visit -snap -proj
+//
+// Output 3:
+//  Relative error of DMD solution (u) at t_final: 2 is 0.0029577152
+//  Elapsed time for solving FOM: 1.989041e-01 second
+//  Elapsed time for training DMD: 1.449013e+00 second
+//  Elapsed time for predicting DMD: 4.555228e-02 second
+//
+// Sample run using snapshotDMD without projecting initial conditions for the
+//  predicition phase.
```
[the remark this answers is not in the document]
ptranq: fixed

lib/algo/SnapshotDMD.h
ptranq: This is a description of what the method is.  It does in fact assume uniform time steps between snapshots.  It is the same as the base DMD class, but so is its behavior.  The only difference between base DMD and snapshot DMD is in "train" and its ability to construct a new set of snapshots via "interpolateToNSnapshots".

lib/algo/SnapshotDMD.h
ptranq: I'm not sure of the original intention of parametricDMD being its own class.  It would be my preference to leave the comments here, so that it is easy to find out what each one does, since this function provides essential functionality to the class.  If noone else agrees, I am happy to ditch them though.

examples/dmd/wave_equation.cpp
ptranq: fixed

examples/dmd/wave_equation.cpp
ptranq: fixed

examples/dmd/wave_equation.cpp
ptranq: fixed

examples/dmd/wave_equation.cpp
```diff
@@ -487,8 +550,16 @@ int main(int argc, char *argv[])
...
+                if(snapshotDMD)
+                {
+                    dmd_u.push_back(new CAROM::SnapshotDMD(u.Size(), dt));
+                }
+                else
+                {
+                    dmd_u.push_back(new CAROM::DMD(u.Size(), dt));
+                }
                 dmd_u[curr_window]->takeSample(u.GetData(), t);
+                dmd_u[curr_window]->takeSample(u.GetData(),t);
```
dylan-copeland, line 562: Should this be called twice?

ptranq: No, I don't believe it should.

lib/algo/SnapshotDMD.cpp
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Copyright (c) 2013-2024, Lawrence Livermore National Security, LLC
+ * and other libROM project developers. See the top-level COPYRIGHT
+ * file for details.
+ *
+ * SPDX-License-Identifier: (Apache-2.0 OR MIT)
+ *
+ *****************************************************************************/
+
+// Description: Implementation of the AdaptiveDMD algorithm.
```
dylan-copeland, line 11: `AdaptiveDMD` -> `SnapshotDMD`

dylan-copeland (approved):

siuwuncheung (approved):

PR #292, opened 2024-06-27
Remove local_tw_csv script

siuwuncheung:
**### (Requires the attention of all folks working on DMD.)**
This PR removes the `local_tw_csv.cpp`, as its functionality is included in `parametric_tw_csv.cpp`.

There are some additional changes in `parametric_tw_csv.cpp`: 
- for the fixed time step case, the DMD default option is reset. 
- fix a segmentation fault when there are multiple testing parameters (thanks @ptranq for catching it)
- the command lines and results are verified and added to the header. 
- the instruction of the MFEM heat conduction database generation is made a little clearer.
- add regression tests for the capabilities


examples/dmd/parametric_tw_csv.cpp
```diff
@@ -10,24 +10,44 @@
 
 // Compile with: make parametric_tw_csv
 //
-// Generate CSV or HDF database on heat conduction with either
-// heat_conduction_csv.sh or heat_conduction_hdf.sh (HDF is more efficient).
+// This script performs serial or time windowing DMD on general datasets in CSV or HDF formats.
```
dylan-copeland, line 13: I think `serial` here means without time windowing, but `serial` may be understood as "single-process". Would it be better to say "DMD with or without time windowing"?

examples/dmd/parametric_tw_csv.cpp
```diff
@@ -10,24 +10,44 @@
...
+// Below is a simple hands-on example, using MFEM heat conduction example to
+// generate either CSV or HDF database to illustrate the file hierarchy and command lines
+// (HDF is more efficient).
 //
 // =============================================================================
 //
+// CSV database generation: heat_conduction_csv.sh
+// HDF database generation: heat_conduction_hdf.sh
+//
+// Local serial DMD command (for HDF version, append -hdf):
```
dylan-copeland, line 24: This works for the CSV version but fails for HDF.

dylan-copeland: Nevermind, it seems I forgot to run the HDF script.

dylan-copeland (approved):

chldkdtn (approved):

PR #295, opened 2024-07-18
Add unit tests to CMake

ckendrick:
This registers unit tests in CMake so tests can be run with `ctest`. This also adds a build option `ENABLE_TESTS` to control whether to build unit tests. Previously, tests were built if googletest was found.


dylan-copeland:
Should the unit tests still work with Google Test as before? Is there a new way to run the tests now?

ckendrick:
Yes, the tests should still function as before with Google test. This just adds a new method for running all tests by using the `ctest` command in the build directory instead of each executable individually (see the changes in `.github/workflows/run_tests/action.yml` for an example).

dylan-copeland (approved):

chldkdtn (approved):

PR #296, opened 2024-08-03
check integer overflow in `Matrix::setSize`

dreamer2368:
So far, `Matrix` does not check if the new size is overflowed. This returns a memory error far downstream of application, which takes a long time to backtrace.

Now `Matrix::setSize` checks if the `new_size` is overflowed at its execution.

### DIFF
--- a/lib/linalg/Matrix.cpp
+++ b/lib/linalg/Matrix.cpp
@@ -132,7 +132,7 @@ Matrix::Matrix(
     else {
         // Check integer multiplication overflow
         if (num_rows > INT_MAX / num_cols)
-            CAROM_ERROR("Matrix::setSize- new size exceeds maximum integer value!\n");
+            CAROM_ERROR("Matrix::Matrix- new size exceeds maximum integer value!\n");
 
         d_mat = mat;
         d_alloc_size = num_rows*num_cols;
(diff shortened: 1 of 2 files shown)


dylan-copeland:
This check is a good improvement, but I think there is a case in the constructor at Matrix.cpp:133, where `setSize` is not called, and the check should be added there as well.

ckendrick (commented):
Thanks for adding this check. This might be a good time to also add this to `Vector::setSize` as well. `Vector` has another case where this check should be added to the constructor at Vector.cpp:69.

dreamer2368:
@dylan-copeland suggestion reflected.

@ckendrick I looked into `Vector` class, and I think it's actually not the case with the same issue. For `Matrix`, the size is determined by multiplication inside the class, while for `Vector` the size is determined simply by a passed input value. We can only check overflow of internal operations. There is no way we can check if the passed `dim` is overflowed or not: a negative overflown value will be caught by `dim > 0`; but a positive overflown value, we cannot tell whether that is proper or overflown.

Also for `Vector` it is easier for users to check if the passed `dim` is a wrong value, while for `Matrix` we need to check the internal `d_alloc_size`.

dylan-copeland (approved):

ckendrick (approved):

PR #298, opened 2024-08-29
Add valgrind run to CI

ckendrick:
This adds a new stage to the CI to run valgrind on the poisson global rom example for informational purposes about potential memory leaks.


chldkdtn:
Are you going to do this only for the Poisson example?

PR #300, opened 2024-09-05
distribute/gather functions for `Vector`

dreamer2368:
This implements `Vector::distribute` and `Vector::gather` analogous to `Matrix` class.

- `Vector::distribute` distributes a non-distributed vector into a locally distributed vector with specified local dimension.
- `Vector::gather` gathers a locally distributed vector into a non-distributed vector.

### DIFF
--- a/lib/linalg/Vector.cpp
+++ b/lib/linalg/Vector.cpp
@@ -637,6 +637,7 @@ Vector::gather()
     delete [] data_offsets, data_cnts;
     d_vec = d_new_vec;
     d_alloc_size = new_size;
+    d_dim = new_size;
 
     d_distributed = false;
 }
--- a/lib/linalg/Vector.cpp
+++ b/lib/linalg/Vector.cpp
@@ -43,9 +43,11 @@ Vector::Vector(
     MPI_Initialized(&mpi_init);
     if (mpi_init) {
         MPI_Comm_size(MPI_COMM_WORLD, &d_num_procs);
+        MPI_Comm_rank(MPI_COMM_WORLD, &d_rank);
     }
     else {
         d_num_procs = 1;
+        d_rank = 0;
     }
 }
 
@@ -74,9 +76,11 @@ Vector::Vector(
     MPI_Initialized(&mpi_init);
     if (mpi_init) {
         MPI_Comm_size(MPI_COMM_WORLD, &d_num_procs);
+        MPI_Comm_rank(MPI_COMM_WORLD, &d_rank);
     }
     else {
         d_num_procs = 1;
+        d_rank = 0;
     }
 }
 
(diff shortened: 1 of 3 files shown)


lib/linalg/Vector.cpp
```diff
@@ -577,6 +584,64 @@ double Vector::localMin(int nmax)
     return v;
 }
 
+void
+Vector::distribute(const int &local_dim)
```
dylan-copeland, line 588: Pass by value?

lib/linalg/Vector.cpp
```diff
@@ -577,6 +584,64 @@ double Vector::localMin(int nmax)
...
+{
+    CAROM_VERIFY(!distributed());
+    CAROM_VERIFY(d_owns_data);
```
dylan-copeland, line 591: Verify that `local_dim >= 0`?

lib/linalg/Vector.h
```diff
@@ -759,6 +759,27 @@ class Vector
...
+     * @brief Distribute this vector among MPI processes,
+     * based on the specified local dimension.
+     * This becomes distributed after this function is executed.
+     *
+     * @pre !distributed()
+     * @pre d_owns_data
+     *
+     * @param[in] local_dim dimension for local MPI rank.
+     */
+    void distribute(const int &local_dim);
```
dylan-copeland, line 772: These functions would be more general if an MPI communicator is an argument, with default value `MPI_COMM_WORLD`.

dylan-copeland: On second thought, `MPI_COMM_WORLD` is used in other places in the class, so the `Vector` constructor should take an optional argument for the communicator. It's not really necessary to do this in this PR, but it's a possibility.

lib/linalg/Vector.h
```diff
@@ -759,6 +759,27 @@ class Vector
...
+
+    /**
+     * @brief Gather all the distributed elements among MPI processes.
+     * This becomes not distributed after this function is executed.
```
dylan-copeland, line 776: Further comment that the data is identical on all processes after this operation?

lib/linalg/Vector.h
```diff
@@ -759,6 +759,27 @@ class Vector
...
 
+    /**
+     * @brief Distribute this vector among MPI processes,
+     * based on the specified local dimension.
+     * This becomes distributed after this function is executed.
+     *
+     * @pre !distributed()
+     * @pre d_owns_data
+     *
+     * @param[in] local_dim dimension for local MPI rank.
```
dylan-copeland, line 770: Comment that `local_dim` must sum to the current dimension over all processes?

lib/linalg/Vector.h
dreamer2368: It was also one of my wish, that libROM parallel linear algebra should support custom MPI communicator. I attempted it a while ago, but it wasn't as simple as I thought, particularly in the interface with scalapack. We can discuss about this more.

For now, in order to be consistent with the current practice throughout the code, I kept `MPI_COMM_WORLD`.

dylan-copeland (approved):

chldkdtn (approved):

ckendrick (approved):

PR #301, opened 2024-09-19
Add Gitlab CI

ckendrick:
This adds a setup to run Gitlab CI on LC Machines.


chldkdtn:
Per PR, we need to trigger regression test in gitlab side somewhat automatic fashion. @ckendrick will look into this.

PR #318, opened 2025-09-29
Update to C++17

dylan-copeland:
Building with MFEM now requires C++17, and the necessary changes are made in this PR.


ckendrick (approved):

examples/prom/de_parametric_maxwell_greedy.cpp
```diff
@@ -91,7 +91,7 @@ bool static_cond = false;
 bool pa = false;
 const char *device_config = "cpu";
 bool visualization = false;
-bool visit = false;
+bool use_visit = false;
```
IdoAkkerman, line 94: removing `using namespace std;` will likely also do the trick. 

I think `using namespace std;`  is considered bad practice, exactly for the reason it causes problems here... I also use it quite often :), just saying..

IdoAkkerman: A good compromise might be:`using namespace std::cout`

According to:
https://www.geeksforgeeks.org/cpp/using-namespace-std-considered-bad-practice/

ckendrick:
Closing since these changes are added in #320.

PR #320, opened 2025-11-21
Rationalize CMake and switch to C++17 to accommodate MFEM

IdoAkkerman:
This PR does three related things -- in order of impact/controversialness(sic):

1. It rationalized/simplifies the CMake routines. 
- In the old routines the MFEM library defaulted to dependencies/mfem even if another mfem was provided
- The mfem is imported as a packages, this simplifies things
- Examples and tests are handled in there own directory, this simplifies the routines
2. Switched to c++23 to accommodate MFEM
 - This created issues with std
 - Removed `using namespace std` from the library and examples
3. Fixed some other (pedantic) compiler errors
 - Added some override statements
 - Made a constructor virtual
 - Separate delete statement for different variables
 - Corrected a string comparison

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,133 +110,26 @@ set(CAROM_HAVE_LAPACK ${LAPACK_FOUND})
 set(CAROM_HAVE_HDF5 ${HDF5_FOUND})
 
 if (USE_MFEM)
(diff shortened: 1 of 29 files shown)


IdoAkkerman:
This PR address issues also tackled in PR #318

ckendrick (approved):
Thanks for making these changes! This looks good to me and works as expected.

dylan-copeland (approved):

